PR #2, opened 2015-09-11
fixed link to SciHub User Guide

Fernerkundung:
SciHub updated the links to the user guide recently. The old link leads to a authentication page with no way of knowing where the actual user guide is located now.


willemarcel:
Merged! Thank you!

Issue #1, opened 2015-08-22
Issues with the certificate when downloading from the Sentinel Hub


h4k1m0u:
Hi,

First of all, thanks for this helpful module.

I experienced an error with the `certificate` when trying to download a specific product from the `Sentinel Hub`.

Here is the command I entered in the shell followed by the error I got:

    $ ~/Scripts/sentinelsat $ sentinel download <user> <password> 5c9746a6-4cac-4344-a86c-8873c487448a
    Downloading 5c9746a6-4cac-4344-a86c-8873c487448a to ./S1A_IW_GRDH_1SDV_20150819T171924_20150819T171949_007335_00A129_A1AC.zip
    Traceback (most recent call last):
    File "/usr/local/bin/sentinel", line 9, in <module>
    load_entry_point('sentinelsat==0.3', 'console_scripts', 'sentinel')()
    File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 700, in __call__
    return self.main(*args, **kwargs)
    File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 680, in main
    rv = self.invoke(ctx)
    File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 1027, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
    File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 873, in invoke
    return ctx.invoke(self.callback, **ctx.params)
    File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 508, in invoke
    return callback(*args, **kwargs)
    File "/usr/local/lib/python2.7/dist-packages/sentinelsat/scripts/cli.py", line 58, in download
    api.download(productid, path)
    File "/usr/local/lib/python2.7/dist-packages/sentinelsat/sentinel.py", line 110, in download
    download(product['url'], path=path, session=self.session)
    File "/usr/local/lib/python2.7/dist-packages/homura.py", line 255, in download
    hm.start()
    File "/usr/local/lib/python2.7/dist-packages/homura.py", line 179, in start
    raise e
    pycurl.error: (60, 'server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none')

The only way to get rid of this error was to open the `homura` module > `homura.py` file > `Homura::curl()` method and make curl not verify the `certificate` of `Sentinel Hub` server (which is not very secure as the download operation will prone be to a `man in the middle` attack):

[...]

PR #3, opened 2015-09-16
get footprints of search results

Fernerkundung:
I added a function to create a geojson FeatureCollection containing all footprint polygons and relevant metadata of a search result. This can also be invoked from the cli with the *-f* option, creating a geojson file of the search result to get a quick overlook of the spatial distribution of the search results. This is especially useful if you are searching in larger areas (think whole countries).

*get_product_info()* now also contains the footprint coordinates in the returned dict and the test for it has been expanded.


Fernerkundung:
14a8042 should solve SSL issues #1 regardless of the OS

Fernerkundung:
14a8042 should be ignored in favour of PR #4

willemarcel:
Merged, thanks! I published a new version of sentinelsat on Pypi.

PR #4, opened 2015-09-17
pass any keyword argument to homura and fix CA errors

### LINKED ISSUE #1: Issues with the certificate when downloading from the Sentinel Hub

citterio:
This makes it possible e.g. to hide the progress bar with `show_progress=False` and other options already supported in the homura version on pypi, and most importantly to close #1 by providing a mechanism to pass the CA certificate path through homura to pycurl (once shichao-an/homura#4 is addressed).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -106,6 +106,24 @@ To download all the results of your query, use:
 
     api.download_all()
 
+The download from <https://scihub.esa.int> will fail if the server certificate 
+cannot be verified because no default CA bundle is defined, as on Windows, or
+when the CA bundle is outdated. In most cases the easiest solution is to
+install or update `certifi <https://pypi.python.org/pypi/certifi>`:
+
+.. code-block:: console
+
+    pip install -U certifi
+
+You can also override the the path setting to the PEM file of the CA bundle using
+the `pass_through_opts` keyword argument when calling `api.download()` or `api.download_all()`:
+
+.. code-block:: python
+
+    from pycurl import CAINFO
+    api.download_all(pass_through_opts={CAINFO: 'path/to/my/cacert.pem'})
+
+
 License
 =======
 
--- a/README.rst
+++ b/README.rst
@@ -77,6 +77,13 @@ If you know the id of the product you want to download, you can download it by u
 
     api.download(<product_id>)
 
+It is possible to hide the progress report, disable resume and auto_retry, and 
+pass any other keyword argument understood by the underlying homura library, e.g.:
+
+.. code-block:: python
+
+    api.download(<product_id>, show_progress=False, max_rst_retries=2)
+
 You can search products by specifying the coordinates of the area and a date interval:
 
 .. code-block:: python
@@ -109,14 +116,15 @@ To download all the results of your query, use:
 The download from https://scihub.esa.int will fail if the server certificate 
 cannot be verified because no default CA bundle is defined, as on Windows, or
 when the CA bundle is outdated. In most cases the easiest solution is to
-install or update `certifi <https://pypi.python.org/pypi/certifi>`:
+install or update `certifi <https://pypi.python.org/pypi/certifi>`_:
 
 .. code-block:: console
 
     pip install -U certifi
 
 You can also override the the path setting to the PEM file of the CA bundle using
-the pass_through_opts keyword argument when calling api.download() or api.download_all():
+the ``pass_through_opts`` keyword argument when calling ``api.download()`` or 
+``api.download_all()``:
 
 .. code-block:: python
 
--- a/README.rst
+++ b/README.rst
@@ -106,7 +106,7 @@ To download all the results of your query, use:
 
     api.download_all()
 
-The download from <https://scihub.esa.int> will fail if the server certificate 
+The download from https://scihub.esa.int will fail if the server certificate 
 cannot be verified because no default CA bundle is defined, as on Windows, or
 when the CA bundle is outdated. In most cases the easiest solution is to
 install or update `certifi <https://pypi.python.org/pypi/certifi>`:
@@ -116,7 +116,7 @@ install or update `certifi <https://pypi.python.org/pypi/certifi>`:
     pip install -U certifi
 
 You can also override the the path setting to the PEM file of the CA bundle using
-the `pass_through_opts` keyword argument when calling `api.download()` or `api.download_all()`:
+the pass_through_opts keyword argument when calling api.download() or api.download_all():
 
 .. code-block:: python
 
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -92,9 +92,10 @@ class SentinelAPI(object):
             ]
         return dict(zip(keys, values))
 
-    def download(self, id, path='.'):
+    def download(self, id, path='.', **kwargs):
         """Download a product using homura's download function. If you don't
-        pass the title of the product, it will use the id as filename.
+        pass the title of the product, it will use the id as filename. Further
+        keyword arguments are passed to homura.
         """
         product = self.get_product_info(id)
         path = join(path, product['title'] + '.zip')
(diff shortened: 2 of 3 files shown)


citterio:
So, homura 0.1.2 with `pass_through_opts` just came up on pypi. With it and this PR, the CA certificate problems in #1 should be solved by doing something like:
```
import pycurl
.
.
.
api.download_all(pass_through_opts={pycurl.CAINFO: 'path/to/cacert.pem'})
```
Where `cacert.pem` contains the trusted certificate of the CA that signed the scihub.esa.int certificate. See e.g. http://curl.haxx.se/docs/caextract.html for resources.

Fernerkundung:
I already implemented this (14a8042) by using `certifi` and tested it with the newest homura version. This way we won't need to hard code a `cacert.pem` but instead can use:

    homura.download(product['url'], path=path, session=self.session, pass_through_opts={pycurl.CAINFO : certifi.where()})

citterio:
@Fernerkundung: certifi is nice but I'm not sure adding it as a dependency and using it out of sight of the user is a good solution here:
- from #1, it seems only somebody needs it to begin with. There must be systems where updating the certificate store used by libcurl is taken care by the system or the user, and we would be interfering with that
- there are networks where the connection goes through a proxy that effectively impersonates the server by replacing the original certificate with its own self signed one that is also installed as a trusted certificated on the local machines. I need to deal with something like this on one of the networks at work and for that, as a user of sentinelsat, I need control over what CA certificate store is used.
- I imagine sentinelsat is mostly used by people setting up long-running downloads jobs and certificates change over time. Unless the user keeps certifi up to date, stale versions of certifi will make a system that worked fine upon install suddenly stop to work some time in the future. As a user I would probably not even realize certifi was there. 

I suggest we import certifi if it is already installed (hopefully then the user knows it is there and to keep it updated) and use it by default if nothing is passed, but give the user the option of overriding it explicitly in the download() and download_all() calls.

Fernerkundung:
@citterio 

Those are valid points I haven't regarded so far. So the hiera

<1> user provides certi

Fernerkundung:
@citterio

Those are valid points I haven't regarded so far. So the hierarchy would be:

1. user provides certificate
2. use the system stored certificate (basically libcurls standard behaviour)
3. use certifi, if it is present on the system

I still think it would be a good idea to alert the user of options 1 and 3 if the connection fails due to certificate errors to spare them some time forging through the libcurl / pycurl documentation.

citterio:
@Fernerkundung
Yes. Alternatively, swapping 3. and 2., as I suppose the only way to fall back to certifi if standard libcurl fails would be to first attempt connecting without any explicit CAINFO and then catch the connection error and retry connecting using the certifi cacert.

citterio:
Now works as discussed above, with priority: 1. user provided path, 2. path to the cacert.pem bundle provided by certifi (if installed), 3. picurl default.

citterio:
@Fernerkundung
Are you suggesting the library should actually raise warnings, or only the command line interface, or is the current info in the readme enough?

Fernerkundung:
@citterio 
Nice work! I'd say the info in the readme is enough since we can reasonably assume that's the first thing soemone turns to when he encounters an error.

citterio:
@Fernerkundung
Agree the error documentation is adequate for the CA issues. A unified treatment of the many different exceptions that can occur would be nice (in the couple of days I've been using scihub I've already collected a number of different errors, including ones that only occur when the ESA server is in maintenance mode and so on) but that is beyond the scope of this PR. 

I'll add a short documentation about what else can now be passed to homura (i.e. the 'pass any keyword argument to homura' title of this PR ) and then it could be good to go.

Fernerkundung:
@citterio
I think a list of errors and their sources in the Readme would be enough for now. As the SciHub is still undergoing constant changes I doubt we'll reach a unified error treatment in the code before the SciHub "leaves beta". Making this even more challenging are the existing discrepancies between the SciHub documentation and it's actual behaviour. For instance the docu says the input polygon may not exceed 200 coordinate points and a bounding box of 10°x10°, when in reality you get perfectly valid results with larger polygons.

There are errors you are unlikely to catch correctly, like the maintenance downtime on 16.09.2015, which curiously returned a `404 error` for me which could also be caused by an outage of my own network.

citterio:
@Fernerkundung
I see what you mean re. SciHub being in flux. I also got the `404` but failed to record it properly. I think it propagated up from requests? Good idea building a list of errors for documentation. It is probably too low level for the readme, but the wiki could be good. @willemarcel?

Glad you mentioned that polygons larger than 10x10 deg. are not officially supported - I hadn't noticed that. I also confirm such oversize polygons seem to work at the moment (mine cover all Greenland). At least, the queries always succeed even when they take substantially longer than average, and they return consistent results. I never checked to see if any file is consistently missing though...

citterio:
This is ready I think. There is no test because the download() method had no tests to begin with and I'm not sure how to test it without providing real SciHub login user and password. This also closes #1.

willemarcel:
Hi, @citterio 

Thanks for your contribution! I'm travelling on vacations now, but I'll merge your requests and update the package in pypi on 28th September.

It's good your suggestion to create a wiki with more documentation too!

Fernerkundung:
@citterio - re: login/password for testing

The login and password for testing purposes are provided as environment variables

```python
environ.get('SENTINEL_USER'),
environ.get('SENTINEL_PASSWORD')
```

willemarcel:
Merged, thanks! I published a new version of sentinelsat on Pypi.

PR #9, opened 2015-10-22
fixed footprint parsing

Fernerkundung:
After ESAs SciHub upgrade the footprint parsing functionality broke. The parsing works again and should be more robust to changes from ESAs side.


willemarcel:
Merged, thanks! I published the version 0.4.2 in pypi.

PR #13, opened 2015-11-26
SciHub server error handling and download retries

Fernerkundung:
Unsucessful attempts to connect to the SciHub, in case it is down or responds in a non expected way, are now handled by most functions and raise Exceptions other than "JSON Decoder Error".

For each scene download the connection is tested beforehand. If no connection was possible it will be retried after 3 minutes. Download starts only after successfull connection to the SciHub.

New tests that mock unsuccessfull connections and pytest fictures for integration tests that depend on a real connection to the scihub.

Minor updates to the installation instructions.


willemarcel:
Merged, Thanks!

Issue #15, opened 2015-11-27
Verify the integrity of the downloaded images


valgur:
Hi,

I really like this utility for its Python and command line interfaces, which I find both to be very clean and well-thought-out. I would love to use it to manage my SciHub downloads but it is currently missing a critical function I need: ensuring that the downloaded images are complete and non-corrupted. Would you consider adding an integrity check after finishing a image download and possibly re-download the image until the check passes successfully (perhaps a limited number of times to prevent an infinite loop)?

Broken downloads are not at all rare in my experience, especially when downloading SLCs. For example, a snippet from a recent downloading session with my personal script:
```
Downloading z:\Sentinel-1a\Latvia\tormimurd\data\S1A_IW_GRDH_1SDV_20141009T160333_20141009T160356_002755_003185_F8C5.zip. Id = c998ec17-c7c6-429f-a2f6-38ff2f26f8f2, Checksum = D317054C7A5E71ADC9EC3A09B85EC15C, Length = 1,576,536,265
0%....10%....20%....30%....40%....50%....60%....70%....80%....90%....100%
Downloaded 1576536265/1576536265 bytes, checksum is D317054C7A5E71ADC9EC3A09B85EC15C.
Downloading z:\Sentinel-1a\Latvia\tormimurd\data\S1A_IW_GRDH_1SDV_20141021T160333_20141021T160356_002930_003530_A0B5.zip. Id = 7dd51ac2-3f26-4b4c-9162-8275ece60149, Checksum = 40824789726D8248F6B55EAFF16BF970, Length = 1,575,458,978
0%....10%....20%....30%....40%....50%....60%....70%....80%....90%....100%
Downloaded 1575458978/1575458978 bytes, checksum is 40824789726D8248F6B55EAFF16BF970.
Downloading z:\Sentinel-1a\Latvia\tormimurd\data\S1A_IW_GRDH_1SDV_20141023T043423_20141023T043448_002952_0035AA_6B85.zip. Id = 427d77d8-9f81-467a-b2a8-e96eae7c36d2, Checksum = C9F1FC5D8B90A8F1126997FA85626CAD, Length = 1,763,514,174
0%....10%....20%....
Downloaded 517355836/1763514174 bytes, checksum is E8A7A9F251199A06DA6FD600A8201C2E.
0%
Downloaded 2718612/1763514174 bytes, checksum is 10E3B673F03AEB28A7945CA57F33ED32.
0%
[...]

PR #16, opened 2015-12-02
MD5 checksum testing after download

Fernerkundung:
Verification of downloaded scenes, as proposed in #15 

The `download`, `download_all` methods and their CLI counterparts now support validation of the downloaded file with the ESA provided MD5 checksum.

- `download` of a single scene raises an error if the check fails
- `download_all` continues the download and returns a list of corrupt scenes that failed the check

The CLI has a new option `-c` which performs the test after each download and creates a text file *corrupted_scenes.txt* with the product ids of the corrupt scenes, if there are any.

Additionaly `sentinel search` now also reports the number of scenes found and their total size.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,7 +5,7 @@ sentinelsat
     :target: http://badge.fury.io/py/sentinelsat
 
 
-Utility pack to search and download Sentinel-1 imagery from `ESA SciHub <https://scihub.esa.int/>.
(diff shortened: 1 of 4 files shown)


valgur:
Excellent! I'll try it out next time I need to do some downloading from SciHub again.

willemarcel:
Sorry for the delay. Merged and published the version 0.5 on Pypi. Thanks!

PR #22, opened 2015-12-30
bugfixes

Fernerkundung:
Some small bugfixes:
- filesize reported by `>sentinel search` has been corrected (MB, GB)
- *corrupt_scenes.txt* now includes filenames as well as IDs of products that failed the md5 check
- md5 is used to test the validity of an already downloaded image instead of filesize
- testing provided api urls for trailing slashes

### DIFF
--- a/sentinelsat/scripts/cli.py
+++ b/sentinelsat/scripts/cli.py
@@ -58,8 +58,8 @@ def search(user, password, geojson, start, end, download, check, footprints, pat
         corrupt_scenes = api.download_all(path, check)
         if len(corrupt_scenes) is not 0:
             with open(os.path.join(path, "corrupt_scenes.txt"), "w") as outfile:
-                for product_id in corrupt_scenes:
-                    outfile.write("%s\n" % product_id)
+                for corrupt_tuple in corrupt_scenes:
+                    outfile.write("%s : %s\n" % corrupt_tuple)
     elif download is True and check is False:
         api.download_all(path)
     else:
--- a/sentinelsat/scripts/cli.py
+++ b/sentinelsat/scripts/cli.py
@@ -63,13 +63,10 @@ def search(user, password, geojson, start, end, download, check, footprints, pat
     elif download is True and check is False:
         api.download_all(path)
     else:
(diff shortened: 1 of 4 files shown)


Fernerkundung:
I'm a bit puzzled by the build failing on Travis as all tests passed on 2 local machines. Most errors seem to revolve around 401 errors.

I suspect the error lies in how Travis handles the secured environment variables for pull requests.

Fernerkundung:
Just confirmed it via the [Travis docs](https://docs.travis-ci.com/user/pull-requests#Security-Restrictions-when-testing-Pull-Requests):

"A pull request sent from a fork of the upstream repository could be manipulated to expose any environment variables. The upstream repository’s maintainer would have no protection against this attack, as pull requests can be sent by anyone with a fork.

For the protection of secure data, Travis CI makes it available only on pull requests coming from the same repository. These are considered trustworthy, as only members with write access to the repository can send them."

willemarcel:
Merged! Thanks. I removed the _url_trail_slash method, because urljoin already concatenates the url parts properly.

Fernerkundung:
@willemarcel 

I thought `urljoin` would handle that, but it doesn't. Regard these examples:

```python
from urlparse import urljoin

print(urljoin("http://scihub.copernicus.eu/apihub", "search?format=json&rows=15000"))
print(urljoin("http://scihub.copernicus.eu/apihub/", "search?format=json&rows=15000"))

>http://scihub.copernicus.eu/search?format=json&rows=15000
>http://scihub.copernicus.eu/apihub/search?format=json&rows=15000
```

In order for sentinelsat to work with `http://scihub.copernicus.eu/apihub` as well as with `http://scihub.copernicus.eu/apihub/` we either need a private function to add the trailing slash or add a preceding slash to all search queries, i.e. `/search?format=json&rows=15000`. 

I tried to go with the "rout of least refactoring" by adding the `_url_trail_slash` function.

willemarcel:
You're right. I put back the method you wrote. Sorry and thank you again!

Fernerkundung:
Thank you for the merge!

Issue #18, opened 2015-12-04
Sentinel-2 support


Fernerkundung:
As of a few hours ago the pre-operational Sentinel-2 hub went live. Luckily the biggest change I can spot so far is the new url (which is already covered with the -u parameter) and some differences in returned metadata.

I think access to Sentinel-2 should be included in 'sentinelsat', especially as it doesn't require a lot of refactoring.

What I tested so far:

- you have to use 'https://scihub.esa.int/s2/' instead of 'https://scihub.copernicus.eu/s2' for the time being or you get a certificate error
- 'sentinel search' and 'sentinel download' behave as expected and I am currently downloading S-2 scenes with it

willemarcel:
I'll need to download Sentinel-2 data soon. We can work to totally support Sentine-2 in sentinelsat.

Fernerkundung:
I'm currently in the midst of writing tests and testing Sentinel-2 support.

@willemarcel Should we expose the Satellite selection and cloud cover in the CLI with their own flags or depend on the user formulating the query terms?

I was thinking something like:

`sentinel search -sentinel2 -cc 30 <user> <password> <geojson>`

This would produce the same result as:

`sentinel search -q "platformname=Sentinel-2,cloudcoverpercentage=[0 TO 30]" <user> <password> <geojson>"

or from Python:

`api.query(platformname="Sentinel-2", cloudcoverpercentage="[0 TO 30]"`

vincentsarago:
hi @Fernerkundung 
Setting the maximum cloud cover seems to be a nice solution (same as landsat-util, https://pythonhosted.org/landsat-util/commands.html) 

Great work,

Vincent

Fernerkundung:
as per #24

- search, parsing and download of Sentinel-2 images works in API and CLI
- CLI has new flags (`-c` for maximum cloud cover, `--sentinel2` and `--sentinel1` to limit search to one constellation)

PR #24, opened 2016-01-13
Sentinel-2 support

Fernerkundung:
Sentinel-2 support as discussed in #18 

The only change necessary to the core logic of `sentinel.py` was the parsing of the returned footprints.

I changed the flags for the CLI to make search and download as easy as possible from there.

- changed `-c` to `--md5`
- `-c` stands for maximum cloud cover (just like in landsat-util)
- `--sentinel1` and `--sentinel2` limit the search to products from one of those satellite (constellations)

Searching for Sentinel-2 scenes over an area with maximum cloud cover of 30% is as easy as `sentinel search -s 20150101 -e 20160101 -c 30 user password roi.geojson`.

In the Python API this would be:

api.query(get_coordinates("roi.geojson"), "20150101", "20160101", "platformname=Sentinel-2", "cloudcoverpercentage=[0 TO 30]")


willemarcel:
Merged. Thank you again!

Issue #26, opened 2016-01-20
Documentation


Fernerkundung:
With new features constantly added to Scihub and therefore also to sentinelsat should the documentation be moved from the Readme.rst to something more structured?

There seem to be two popular options for this:

1. Use Sphinx or manually create the docs and use [Read The Docs](https://readthedocs.org/).
2. Create Documentation for each module with examples (e.g. like [rasterio](https://github.com/mapbox/rasterio/tree/master/docs)).

So far a number of the functions are not documented so users of the API can't use it to it's full potential. This way we could also clean up the Readme.rst to include a quick introduction and a separate documentation for more in-detail examples.

@willemarcel your thoughts on this?

pominville:
It would also be nice to have a version history that explains additions/corrections without having to look into the code.

Fernerkundung:
[![Documentation Status](https://readthedocs.org/projects/sentinelsat/badge/?version=latest)](http://sentinelsat.readthedocs.org/en/latest/?badge=latest)


As of version 0.6.4 the outline for a concise online documentation is done. You will be able to find the docs on ReadTheDocs.org.

Now the documentation needs to be moved from the README to the Sphinx docs inside the `docs/` folder.

Fernerkundung:
Should be handled with #38

PR #27, opened 2016-02-03
Documentation: ReadTheDocs - Sphinx setup

Fernerkundung:
This PR would be the basis for solving #26 

A Sphinx documentation with a webhook to ReadTheDocs would allow us to let RTD handle the versioning, creation and hosting of the documentation.

We would only need to shift the current (incomplete) documentation from the README to the docs/ Sphinx directory.

The first setup currently looks like this: http://sentinelsat.readthedocs.org


willemarcel:
Merged, thanks!

PR #30, opened 2016-04-07
Sphinx setup + Docs

Fernerkundung:
Sphinx is now setup with autodoc/numpydoc. This should enable everyone to contribute to the docstrings which are then automatically incorporated into the Sphinx docs.

The preview is available at: http://sentinelsat.readthedocs.org

After merge the [ReadTheDocs Webhook](http://read-the-docs.readthedocs.org/en/latest/webhooks.html) can be added to the project webhooks and the documentation will be updatet on every PR. 

adresses #26. Still missing some function docstrings as well as a changelog.

After the merge we can de-clutter the README.rst to remove duplicate information and make it a *readme* again instead of a *documentation*.


Fernerkundung:
@willemarcel anything in this that needs improvement?

willemarcel:
Merged. Sorry for the delay.

Fernerkundung:
Thanks. I changed the location so the docs now get built from master.

Only thing missing now is adding the webhook so they get rebuild automatically. Either from the built in github hooks or with travis by `curl -X POST https://readthedocs.org/build/sentinelsat`

PR #33, opened 2016-05-31
search by acquisition date instead of ingestion date

### LINKED ISSUE #32: -s option reads ingested not acquired date
Using the -s option including a date limits the download to the time period where the image was uploaded to the hub, not when it was acquired.

`sentinel search -d -s 20151219 -c 30 --md5 username password roi.geojson`

> Search and download all Sentinel-2 products with less than 30% cloud cover **acquired** since 2015-12-19. 

E.g: using -s 20160525 will download data aquired from 2015 which was just added (ingested) to the Sentinel-2 hub this month.

Fernerkundung:
fixes #32

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -23,9 +23,9 @@ Search areas are provided as GeoJSON polygons, which can be created with
 If you do not specify a start and end date only products published in the last
 24 hours will be queried.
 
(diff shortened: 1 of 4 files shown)


willemarcel:
@Fernerkundung Thanks for the PR. The tests failed because I have a problem in my scihub account. I'll try to solve it soon.

Issue #36, opened 2016-07-08
Support for large queries


yuriiz:
sentinelsat currently can not perform search if query length is more than some limit. For example querying this geojson:
[ETH_adm0.zip](https://github.com/ibamacsr/sentinelsat/files/353595/ETH_adm0.zip)

results in:

```
u@m:~$ ~/.local/bin/sentinel search user password ETH_adm0.geojson 
Error: API returned unexpected response 414 .
Traceback (most recent call last):
  File "/home/u/.local/bin/sentinel", line 9, in <module>
    load_entry_point('sentinelsat', 'console_scripts', 'sentinel')()
  File "/home/u/.local/lib/python2.7/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/home/u/.local/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/home/u/.local/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/u/.local/lib/python2.7/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/u/.local/lib/python2.7/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/home/u/sentinelsat/sentinelsat/scripts/cli.py", line 99, in search
    for product in api.get_products():
  File "/home/u/sentinelsat/sentinelsat/sentinel.py", line 130, in get_products
    raise ValueError('API response not valid. JSON decoding failed.')
ValueError: API response not valid. JSON decoding failed.
u@m:~$ 
```

(414 is Request URI too long)

Fernerkundung:
@willemarcel This can be closed as it was merged with 3646a0e

PR #37, opened 2016-07-08
Support for large queries #36

yuriiz:


Fernerkundung:
Nice catch! Could you adapt the tests so they pass locally? I think changing the mock adapter from GET to POST and the expected formated URL should be pretty straightforward.

yuriiz:
Updated mocks but tests still fail.

Fernerkundung:
The travis tests fail if anyone without write permissions to the repo commits. All tests of your branch are passing locally on my machine. Great job :+1: 

This is looking good to go @willemarcel

PR #39, opened 2016-08-09
Improve downloading, added raw_query(), misc.

valgur:
A bunch of small changes.

#### Highlights
- Added `raw_query()` to the API. This allows the SciHub's full-text search to be accessed directly, which allows for a lot more flexibility, if necessary. Did not add it to the CLI yet, though.
- MD5 checking in `download()` for existing files did not work correctly. Incomplete files were erroneously checked for MD5 and deleted. Now the file size is checked first and only after that the MD5 verified if the new parameter `check_existing` is enabled.
- Added product info to `download()`'s return value.
- Added a progress bar for MD5 calculation.
- Made `download_all()` considerably more robust. Downloading is retried until it succeeds or `max_attempts` is exceeded.
- Added unit tests for `download()` and `download_all()`. Test coverage is now 94% for the API code.


valgur:
The motivation behind these changes was mostly to allow `sentinelsat` to be used for maintaining an archive of images by periodically running something like
```python
from sentinelsat.sentinel import SentinelAPI
query = 'IW AND (GRD OR SLC) AND footprint:"Intersects(POLYGON((21.782143554687 57.495711986813,28.220131835936 57.495711986813,28.220131835936 59.682121552603,21.782143554687 59.682121552603,21.782143554687 57.495711986813)))"'
api = SentinelAPI('user', 'password')
api.query_raw(query)
api.download_all('.', checksum=True, check_existing=False)
```
Also, by setting `check_existing` to True it is now possible to verify existing products' integrity, if necessary.

willemarcel:
Seems interesting! I'll verify and merge it soon!

Fernerkundung:
@willemarcel looks like @valgur is missing in the contributors list.

PR #43, opened 2016-09-19
Ignore unit tests requiring SciHub credentials when Travis CI is run from a pull request

valgur:


valgur:
Alternatively, I suppose we could simply use a non-secret login for the Scientific Data Hub. It would not violate the stated [Terms and Conditions](https://scihub.copernicus.eu/twiki/do/view/SciHubWebPortal/TermsConditions) and it is hard to imagine that anyone would have be interested in stealing this, given how easy it is to register an account over there.
What do you think?

willemarcel:
Thanks @valgur for all your great contributions!!! 

I think it's not a good idea to publish a password. ESA emailed me sometime ago because my user was downloading too much data and making repeated requests.

PR #53, opened 2016-10-14
Wrong exception is being caught in download()

valgur:
`download()` is currently catching the wrong exception and does not wait in case of API errors.


willemarcel:
Thanks, @valgur

PR #55, opened 2016-10-25
Missing RTD requirements

Fernerkundung:
Some modules have been missing from the ReadTheDocs requirements (html2text, tqdm). This causes the build to succeed but the API documentation stays empty.

### DIFF
--- a/docs/readthedocs-requirements.txt
+++ b/docs/readthedocs-requirements.txt
@@ -2,3 +2,4 @@ requests
 click
 geojson
 numpydoc
+html2text
--- a/docs/readthedocs-requirements.txt
+++ b/docs/readthedocs-requirements.txt
@@ -3,3 +3,4 @@ click
 geojson
 numpydoc
 html2text
+tqdm


valgur:
Ah, thanks for catching that. I was not aware of this additional requirements file.

Fernerkundung:
@valgur To be fair, I didn't document it properly. I will try to find a way for RTD to use the original requirements.txt so we don't have to maintain two requirement files.

valgur:
If it is used like a normal requirements.txt file, then `-r ../requirements.txt` might work.

Fernerkundung:
That works, but then we would need to include `numpydoc` into the general requirements file - which is unnecessary.

valgur:
Sorry, I should have been more clear. I meant that as the content of the `readthedocs-requirements.txt` file. You can specify recursive dependencies with `-r`. With `numpydoc` included the contents would be
```
numpydoc
-r ../requirements.txt
```

Fernerkundung:
Great tip! Just tried it on my branch and works like a charm.

Issue #57, opened 2016-10-26
OpenSearch limits number of rows


Fernerkundung:
Bad [news from ESA](https://scihub.copernicus.eu/news/News00111):

>many scripts using the OpenSearch API have been attempting to retrieve extremely large numbers of results from queries via use of the OpenSearch ‘rows' parameter. A new behaviour has been implemented in the data hub software to protect for this and allow a maximum number of rows to be returned in a single query, set at 100 for the moment.

At the moment we query [up to 1500 rows](https://github.com/ibamacsr/sentinelsat/blob/0ed829b211811cf16898acd7c6726bfb853b24cd/sentinelsat/sentinel.py#L132). Meaning we will get `HTTP Error 500` all the time.

To absolve this a new logic is required that iteratively queries the Scihub for results 1-100, 101-200, 201-300, etc. until no more results are returned.

Fortunately this can be done with the OpenSearch `step` parameter:
>Results sets over the maximum can be obtained through paging of from different start values via use of the OpenSearch 'start' parameter.

j08lue:
Are you working on this already, @Fernerkundung? Else I could help.

Fernerkundung:
I won't be able to tackle this for the next two weeks. Would be great if you could chime in @j08lue.

Fernerkundung:
Solved with #60

Issue #52, opened 2016-10-03
Download doesn't work anymore


Rot90:
Hello,
I used sentinelsat until 4 days ago when this error showed up: 

Traceback (most recent call last):
  File "c:\users\crossi\appdata\local\continuum\anaconda2\lib\runpy.py", line 17
4, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "c:\users\crossi\appdata\local\continuum\anaconda2\lib\runpy.py", line 72
, in _run_code
    exec code in run_globals
  File "C:\Users\crossi\AppData\Local\Continuum\Anaconda2\Scripts\sentinel.exe\_
_main__.py", line 9, in <module>
  File "c:\users\crossi\appdata\local\continuum\anaconda2\lib\site-packages\clic
k\core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "c:\users\crossi\appdata\local\continuum\anaconda2\lib\site-packages\clic
k\core.py", line 696, in main
    rv = self.invoke(ctx)
  File "c:\users\crossi\appdata\local\continuum\anaconda2\lib\site-packages\clic
k\core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "c:\users\crossi\appdata\local\continuum\anaconda2\lib\site-packages\clic
k\core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "c:\users\crossi\appdata\local\continuum\anaconda2\lib\site-packages\clic
k\core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "c:\users\crossi\appdata\local\continuum\anaconda2\lib\site-packages\sent
inelsat\scripts\cli.py", line 83, in search
    api.query(get_coordinates(geojson), start, end, **search_kwargs)
  File "c:\users\crossi\appdata\local\continuum\anaconda2\lib\site-packages\sent
inelsat\sentinel.py", line 130, in query
    self.query_raw(query)
  File "c:\users\crossi\appdata\local\continuum\anaconda2\lib\site-packages\sent
inelsat\sentinel.py", line 138, in query_raw
    _check_scihub_response(self.content)
  File "c:\users\crossi\appdata\local\continuum\anaconda2\lib\site-packages\sent
inelsat\sentinel.py", line 89, in _check_scihub_response
    raise SentinelAPIError(response.status_code, code, msg, response.content)
[...]

PR #59, opened 2016-10-27
remove _url_trail_slash

j08lue:
`_url_trail_slash` is broken as discussed in https://github.com/ibamacsr/sentinelsat/issues/52#issuecomment-256567082 .

But as far as I can see, we do not need to take care whether the URL has a trailing slash or not, since we use `urljoin` for all subsequent URL manipulations.

So this PR removes `_url_trail_slash` entirely.


Fernerkundung:
This PR solves #52. The remaining failed tests are related to #57.

PR #60, opened 2016-10-27
Limit rows and query in multiple iterations

j08lue:
This PR addresses issue #57.

The solution required some restructuring of the api methods:

1. A new method `load_query` combines both, the execution of a query against the OpenSearch API (formerly done in `query_raw`, which was removed) and the parsing and evaluation of the JSON response (formerly done in `get_products`).

1. The attribute `url` is now an actual attribute, set by a new function, `format_url`, which generates the SciHub url with a `start` parameter (the starting row for paging).

1. The `products` attribute is now a list from the beginning (`init`) that gets extended by every call to `query` or `load_query`. This change is not strictly necessary (`products` could also be reset for every `query` call), but one I find logical: You can execute several queries and the api instance will store the products, e.g. for a single subsequent call to `download_all`.

None of these changes affect the CLI and for users who just used the `query` method, nothing changed.

The tests are changed accordingly and I added one for a large query that returns over 100 products to show that it works.

NB: Some credit for this code goes to @radosuav.

### DIFF
--- a/tests/test_mod.py
+++ b/tests/test_mod.py
@@ -111,23 +111,6 @@ def test_set_base_url():
 
 
 @pytest.mark.fast
-def test_trail_slash_base_url():
-    base_urls = [
-        'https://scihub.copernicus.eu/dhus/',
-        'https://scihub.copernicus.eu/dhus'
-    ]
-
-    expected = 'https://scihub.copernicus.eu/dhus/'
-
-    for test_url in base_urls:
-        assert SentinelAPI._url_trail_slash(test_url) == expected
-        api = SentinelAPI("mock_user", "mock_password",
-            test_url
-        )
-        assert api.api_url == expected
-
-
-@pytest.mark.fast
 def test_get_coordinates():
     coords = ('-66.2695312 -8.0592296,-66.2695312 0.7031074,' +
               '-57.3046875 0.7031074,-57.3046875 -8.0592296,-66.2695312 -8.0592296')
--- a/tests/test_mod.py
+++ b/tests/test_mod.py
@@ -22,8 +22,8 @@ _large_query = dict(
         end_date=datetime(2015, 12, 31))
 
 _api_auth = dict(
-        user=environ['SENTINEL_USER'],
-        password=environ['SENTINEL_PASSWORD'])
+        user=environ.get('SENTINEL_USER'),
+        password=environ.get('SENTINEL_PASSWORD'))
 
 _api_kwargs = dict(_api_auth,
         api_url='https://scihub.copernicus.eu/apihub/')
--- a/tests/test_mod.py
+++ b/tests/test_mod.py
@@ -72,12 +72,10 @@ def test_api_query_format():
     now = datetime.now()
     query = api.format_query('0 0,1 1,0 1,0 0', end_date=now)
     last_24h = format_date(now - timedelta(hours=24))
-    assert api.url == 'https://scihub.copernicus.eu/apihub/search?format=json&rows=15000'
     assert query == '(beginPosition:[%s TO %s]) ' % (last_24h, format_date(now)) + \
                     'AND (footprint:"Intersects(POLYGON((0 0,1 1,0 1,0 0)))")'
 
     query = api.format_query('0 0,1 1,0 1,0 0', end_date=now, producttype='SLC')
-    assert api.url == 'https://scihub.copernicus.eu/apihub/search?format=json&rows=15000'
     assert query == '(beginPosition:[%s TO %s]) ' % (last_24h, format_date(now)) + \
                     'AND (footprint:"Intersects(POLYGON((0 0,1 1,0 1,0 0)))") ' + \
                     'AND (producttype:SLC)'
--- a/tests/test_mod.py
+++ b/tests/test_mod.py
@@ -11,6 +11,23 @@ import requests_mock
 from sentinelsat.sentinel import (InvalidChecksumError, SentinelAPI, SentinelAPIError, convert_timestamp, format_date,
                                   get_coordinates, md5_compare)
 
+_small_query = dict(
+        area='0 0,1 1,0 1,0 0',
+        initial_date=datetime(2015, 1, 1),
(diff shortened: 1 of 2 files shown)


j08lue:
I think you need to [set environment variables for Travis CI](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings).

All tests pass when I run `python -m pytest tests` with `SENTINEL_USER` and `SENTINEL_PASSWORD`.

Fernerkundung:
Great work @j08lue and @radosuav!

The environment variables for Travis CI are set but as per the docu you linked:
> Encrypted environment variables are not available to pull requests from forks due to the security risk of exposing such information to unknown code.

All tests pass locally on 2.7 and 3.5 - this should be ready to merge.

tests/test_mod.py
```diff
@@ -11,6 +11,23 @@
...
+        area='0 0,1 1,0 1,0 0',
+        initial_date=datetime(2015, 1, 1),
+        end_date=datetime(2015, 1, 2))
+
+_large_query = dict(
+        area='0 0,0 10,10 10,10 0,0 0',
+        initial_date=datetime(2015, 1, 1),
+        end_date=datetime(2015, 12, 31))
+
+_api_auth = dict(
```
Fernerkundung: instead of `environ` I suggest to get the credentials with `environ.get`. This way we don't get an exception on TravisCI when a PR is created from a fork.

j08lue: Good idea, so you mean like `environ.get('USERNAME', 'some_default')`? But what should `some_default` be? If it is not a valid user name, many of the tests will still fail.

Fernerkundung: Actually just `environ.get('USERNAME')`. The scihub integration tests are [expected to fail](https://github.com/ibamacsr/sentinelsat/blob/0ed829b211811cf16898acd7c6726bfb853b24cd/.travis.yml#L17) on PRs from forks.

With Travis there is no way to securely provide credentials for the integration tests unless the PR comes from someone with write access to the repository.

j08lue: cool! Travis rocks. :-) Let's see if this works then.

andrejreznik:
I clone your [branch](https://github.com/DHI-GRAS/sentinelsat) and I got same error

> **message** _Request processing failed; nested exception is java.lang.IllegalArgumentException: parameter `rows` exceeds the maximum value (100)_

j08lue:
@andrejreznik did you checkout the `rows` branch? The `master` branch does not have these changes.

andrejreznik:
@j08lue sure.

j08lue:
@andrejreznik your code is good. I think it could be a server problem. I have over the past hours sometimes gotten errors and sometimes not - for identical queries. So I think the queries are formatted correctly. Please keep me posted whether your code at some point works, then we know that it was just a(nother) server problem.

tests/test_mod.py
valgur: I think it would be a good idea to check if the tests marked with `@pytest.mark.scihub` are enabled and fail fast if the environment variables are not set. I haven't looked into how this might be done but I assume it should not be too difficult.

j08lue: Good idea, @valgur. Maybe we could use [`skipif`](http://doc.pytest.org/en/latest/skipping.html#id1) for that, e.g. testing whether `_api_auth is None`. However, this would mean that these tests are also (silently?) skipped if the environment variables are not declared by mistake.

Fernerkundung: @j08lue `pytest` will report skipped/xfailed tests as well, so they won't really fail silently. Still it would be a good idea to point this out in a small section of the readme aimed at developers.

valgur:
Thanks a lot for the changes! They are a definite improvement to the current querying system. Some thoughts (some of which are more related to `sentinelsat` in general):

1. Changing the `load_query()` to return a generator instead of a full list might be a good idea. As I see it, the motivation on the SciHub side for limiting the maximum query sizes was to limit spikes caused by large queries. This would spread out the queries made when iterating over the query results in `download_all()`, for example. There is one minor problem, though. The SciHub's OpenSearch API provides the number of rows a full query result would contain but Python generators don't support `len()`, so that information would need to be provided via some other means.
2. I would rename the `max_query` to `page_size` (since we are doing pagination) and change the `max_query` to actually limit the number of items returned. Again, to reduce stress on the SciHub server if we actually do just need a limited number of query results.
3. I like the idea of concatenating consecutive query results but I think should only be done in the CLI. More generally, I dislike the current statefulness of the `sentinelsat`'s Python API and don't find this to be a particularly good choice for an API design. I would much rather prefer the `sentinelsat` to return the query results and add an extra positional parameter `products` to `download_all()`.
4. I preferred the `query_raw()` over `load_query()` which sounds too similar and redundant to `query()` to me. Could be named something else as well, but either way, I think a function that exposes the text-search format of the SciHub server should be a first-class citizen in the API, since it is much more powerful, if necessary, than the main `query()` function. For example, it allows for any logical statements that include `OR` and parentheses to be used. I also occasionally find `footprint:Contains()` to be useful instead of `footprint:Intersects()`. I have been intending to add the raw text query support to the CLI as well but have not found time for that yet. 

Comments, thoughts? I think this pull request is good as it is, though, and these could also be addressed in another PR.

willemarcel:
Thanks, @j08lue! I just published the version 0.7.1 in Pypi.

j08lue:
@valgur Those are all very good considerations, and I'd support all of them:

1. Turn the query into a generator in order to further distribute the server stress, (perhaps with a method to load the full `list(iter)` of products).
2. Renaming and clarifying the paging kwargs.
3. Reduce statefulness - let `query` return the products (or a generator) that the user must pass to `download_all`. This makes the interaction more transparent and is useful for example for splitting up the download into chunks.
4. Expose a `query_raw` method again.

We should move this discussion to a new issue and hear the more seasoned contributors to this project.

Fernerkundung:
@j08lue would you mind opening a new issue, so we can keep this discussion ongoing as a WIP?

PR #68, opened 2016-11-21
Correctly handle missing trailing slash in api_url

valgur:


Fernerkundung:
Good catch. The method `_url_trail_slash` was removed in an [earlier PR](https://github.com/ibamacsr/sentinelsat/pull/59/commits/d7ed9d2d1f13376de07f6a29fd5df2e6607377da) because it was broken.

However we forgot that `urljoin('https://scihub.copernicus.eu/apihub', 'search?format=...')` returns `'https://scihub.copernicus.eu/search?format=...'`

sentinelsat/sentinel.py
```diff
@@ -122,7 +122,7 @@ class SentinelAPI(object):
     def __init__(self, user, password, api_url='https://scihub.copernicus.eu/apihub/', max_rows=100):
         self.session = requests.Session()
         self.session.auth = (user, password)
-        self.api_url = api_url
+        self.api_url = api_url if api_url[-1] == '/' else api_url + '/'
```
j08lue: Why not use `if api_url.endswith('/')` here? Same same but cleaner?

valgur: Sure, that's a bit better. I'll change that.

PR #69, opened 2016-11-21
Use vcrpy to record and replay queries made during testing

valgur:
As you've probably noticed, SciHub is down extremely frequently. I thought it might be a good idea to fully mock the queries made to it and not necessarily depend on it for testing. I found a nice library called [vcrpy](https://github.com/kevin1024/vcrpy) that tackles this task very well. It only required some minor configuration and adding decorators to tests.
The benefits are obvious: the tests become deterministic, preventing the build from needlessly "failing", and most of the unit tests will be able to run on pull requests.

I configured it to generate the "cassettes" containing the interactions with SciHub on the first run and then use these cassettes on any runs thereafter. If the queries made by the test change and are missing from the cassette, an exception will be thrown and the cassette file has to be deleted to re-record the session.

I made sure that any sensitive information is not recorded (that is, no cookies are saved).

I guess it would be nice to also have options when running the tests to ignore the recordings and interact with SciHub directly and possibly re-record the sessions. These could be added later.


Fernerkundung:
Great addition. Just a minor comment: 

Can you please include the creation of the cassettes and maybe 2 sentences about the use of VCR into the docs and the README under testing? If I understood it correctly re-creating the cassetes would be as easy as deleting them and re-running the tests (with a working scihub ofcourse)?

`py.test -v -m "homura"` are our only full integration tests at the moment. Having a flag to run the other tests with or without the use of cassettes would be nice but not strictly necessary as one can simply delete the cassetes and have the tests then run directly against sci-hub. As long as it's documented I think we're fine.

willemarcel:
Thanks, @valgur !

Issue #65, opened 2016-11-01
Reduce statefulness of `SentinelAPI`


j08lue:
This is also a discussion from PR #60.

Currently, `api.query` stores the results in the `api.products` attribute, which `api.download_all` subsequently accesses. I agree with @valgur in [his comment](https://github.com/ibamacsr/sentinelsat/pull/60#issuecomment-256920814) that this feels like a bad design choice, because the state of the `SentinelAPI` instance becomes very important for the whole flow `query > download`. What if we want to perform another query with the same API instance, while a download is in progress? Would this alter the list that the `download_all` routine is at the same time iterating over?

A solution would be to have `query` return the results and the user pass them to `download_all`. Another benefit of this solution would be that it becomes cleaner to filter the query results before downloading. For example, if I want to download my products grouped by date, I currently have to modify the `api.products` list before calling `download_all` but keep the original list.

A change is easy to implement but changes the way users have to deal with the API. Any opinions on whether we should pull this through?

willemarcel:
@valgur @j08lue I agree that we can add another argument to `download_all` so we could define what we want to download with more flexibility. Remember that the Sentinel API is very restricted in terms of parallel downloads and queries. I prefer to avoid executing two actions at the same time.

Fernerkundung:
I'm a firm +1 on the ability to being able to filter the results of `query` before passing them on to `download_all`. 

@willemarcel is right that we should not allow more than two interactions with the SciHub at the same time. ESA states, that it limits the dowloads to two concurrent but I found that 3 simultaneous searches result in random errors as well. So any filtering of the `products` should preferrably happen without initiating new queries.

jamesemwheeler:
Reposting relevant post and response from me and @j08lue from #63:

> I'm using sentinelsat to produce Sentinel 1 footprint coverage from a multi-polygon geojson. Each query command I run contains the results of the previous one, even though not all queries exceed 100 total results. I'm not using this library to download imagery from the scihub data repository so (I imagine) there shouldn't be an issue with overloading their servers.
> 
> It's either a problem with my code (below), the apihub servers holding my query (does that happen?), or the latest update to the query/load_query function. As you can probably see I'm not really equipped to alter your original code. If other people are reporting this perhaps it would be worth finding out what's happening.
> 
> Regards,
> 
> James
> 
> Sample code below - apologies for messiness.
> 
> ```
> import sentinelsat
> import json
> 
> from sentinelsat import sentinel
[...]

Issue #63, opened 2016-11-01
Revert some changes to SentinelAPI from last PR


j08lue:
@valgur, in the last PR, there were a few changes to `SentinelAPI` you wanted to revert. I think [you made good points](https://github.com/ibamacsr/sentinelsat/pull/60#issuecomment-256920814), so here they are for further discussion:

> 2. I would rename the <code>max_query</code> to <code>page_size</code> (since we are doing pagination) and change the <code>max_query</code> to actually limit the number of items returned. Again, to reduce stress on the SciHub server if we actually do just need a limited number of query results.</li>
> 4. I preferred the <code>query_raw()</code> over <code>load_query()</code> which sounds too similar and redundant to <code>query()</code> to me. Could be named something else as well, but either way, I think a function that exposes the text-search format of the SciHub server should be a first-class citizen in the API, since it is much more powerful, if necessary, than the main <code>query()</code> function. For example, it allows for any logical statements that include <code>OR</code> and parentheses to be used. I also occasionally find <code>footprint:Contains()</code> to be useful instead of <code>footprint:Intersects()</code>. I have been intending to add the raw text query support to the CLI as well but have not found time for that yet. </li>

Fernerkundung:
+1 on both:

2. could be usefull to implement something like 'sentinel search ... -head 10' or 'tail' if someone wants to see some results of the queried area without limiting the query too much. I find myself sometimes to expand the time window unnecessarily and then just looking at the first 5 scenes returned.

3. Never used queries in that way but I do see the added value of this. Adds flexibility for power users without limiting the ease of use for anyone else.

j08lue:
I just went about implementing `1.` but got stuck on how to implement the `head`/`tail` or `max_query` limit. 

We could execute the full query and then only return the subset the user asks for. If we go ahead with #65, this will be very straightforward: The user anyways gets the full list of products from the query and can do with it what she wants - subset, filter... The same we could do in the CLI with `head`/`tail` options.

If we want to limit the actual query/page iterations with `max_query`, things get more complicated. What if `max_query` is not a multiple of `page_size`? Should we load the full last page and then still limit the returned number of points? I found that this adds a lot of code to the API functions.

We could also expose `start_row` as an argument to `query`, then savvy users could choose e.g. to start the query on page 3 in steps of 10 rows (getting the `tail` of the products).

[...]

PR #70, opened 2016-11-24
Added function to return ordered products

lucadelu:


Fernerkundung:
Good idea Luca, but I see some issues with it:

1. What value does order="date" add? The results are returned by date allready.
2. What happens if I try to order Sentinel-1 data by cloud-cover `sentinel search --sentinel1 -o cloudcoverpercentage`?
3. A unit-test would be nice to see if this works as intended when we extend the order options. Also to test edge cases.

Overall this is a nice idea but to me feels like something power users would achieve with the API anyway. Sorty by cloudcover in the command line is nice - but that wouldn't limit the downloads the user starts so they have to sift through the results manually as well, which is just short of sorting everything in a text-editor or using the Python API.

lucadelu:
Hi Kersten, 
the answers to your question

1. The result by date is in inverse order than the normal output, from the older to the newer
2. It fail right now, I just discovered it creating the unit-test
3. I'm working on them ;-)

You are right, but maybe this tool will be used not only by power users but also from scientist without big informatics skills.

valgur:
An `orderby` parameter is a good addition, I agree. Is there a reason for not using the SciHub API parameter `$orderby` described in https://scihub.copernicus.eu/userguide/5APIsAndBatchScripting? A minor benefit of doing it on the client side is the possibility to get different orderings without running multiple queries on the server.

j08lue:
@lucadelu Your `--orderby` click option should use [`click.Choice`](http://click.pocoo.org/5/options/#choice-options).

lucadelu:
@valgur thanks a lot to showing me the `$orderby` parameters, right now I will continue on the client side as you said is possible to get different orderings without running multiple requests (I'm still improving it to work correctly with Sentinel-1)

lucadelu:
@j08lue done it ae01189

Fernerkundung:
@valgur @lucadelu 

How would you implement the local filtering without querying SciHub again? That means that the results of the query as well as the query itself need to be cached as we only want to change the order when the query is the same. What happens if the user wants to change the order after a while but there have been new products published that fit the query. How would we decide when to query again and when to just sort locally. This decision becomes very complex, fast.

I'm much more in favour of implementing `-o` as a shortcut to `$orderby` (great tip btw @valgur - didn't know about that one), even though this means potentially more load on the server. This could also be beneficial if for instance a user wants to find the 10 least cloud covered scenes with something like `-o cloudcover -limit 10 <user> <pw> huge_polygon.geojson` (with something like a `head` implementation discussed in #63 ) which would translate in a query that leaves the ordering up to SciHub that just returns 10 results, instead of serving all results in a potentially huge query which we then sort locally.

tl;dr I'd say we use `$orderby`, unless we find something that can only be done locally.

valgur:
One more point for using server-side `$orderby`: we likely want to get rid of `get_products()` in the near future, see #65.

lucadelu:
@Fernerkundung the local filter is already implemented and it order the result of `query` call inside `get_products` function so we don't need to recall the `query` function.

I will implement the `$orderby` parameter to use it in the `sentinel` script, but I would like to maintain my local solution for power users that they use the API and not the script. What do you think?

If in the future `get_products()` will be removed we could just move `_order_by` function somewhere else

Fernerkundung:
@lucadelu I'm +1 on implementing a shortcut to access `$orderby` in the CLI as well as the API. I just looked it up - when no parameter is set it seems to be automatically ordered by ingestionDate, not beginPosition so we should change that anyway.

I still don't see the point in implementing local filter functions and then expose them as class methods. This gets complicated fast and doesn't add a lot of value (especially for power users). The complexity can be allready seen in the switch/case statements we would need to filter illegal filter functions (no cloud cover in Sentinel-1, polarisationmode is valid only for Sentinel-1, etc.) - this is only going to get worse the more Sentinels and possible keywords there are.

Local filters are far easier to implement by power users (if they need them, instead of `$orderby`). Something like this to sort by cloud cover:

```python
from sentinelsat.sentinel import SentinelAPI, get_coordinates
api = SentinelAPI("user", "pw")
result = api.query(area=get_coordinates('munich.geojson'), initial_date="20161101", \
                   end_date="20161120", platformname="Sentinel-2")
result.sort(key=lambda x: float(x["double"]["content"]))
```

This leaves us with a much cleaner and easier to maintain codebase, since all the heavy lifting is done by the SciHub while still enabling power-users to filter at their will. We could however include local filtering procedures into the docs - we could need some more examples.


tl;dr +1 on exposing `$orderby`, -1 on local filters

valgur:
If we do want to support sorting and filtering of the query response, then I think a better solution might be to add a convenience function that converts the response to a `(geo)pandas` `(Geo)DataFrame`. That would allow you to slice and dice the list of products conveniently in any imaginable way. Maybe also adjust the `api.download_all()` to also support `df.to_dict()` in that case.

lucadelu:
@Fernerkundung ok I will implement `$orderby` in a new branch and close this pull request

lucadelu:
@valgur are you sure that `@orderby` is supported by https://scihub.copernicus.eu/apihub/search? i did some test but I'm not able to work with it, it seems is possible to use only with odata protocol. Maybe I'm missing something...

lucadelu:
@valgur @Fernerkundung do you have idea is `@orderby` is supported by https://scihub.copernicus.eu/apihub/search?

valgur:
Hm, thanks for pointing that out. I did not notice that the `$orderby` is a parameter only for the OData interface on SciHub. It does not seem to have any effect on the OpenSearch API.
Since server-side sorting is not supported, then I guess we should consider adding some kind of sorting on the client side.

Fernerkundung:
@lucadelu You're absolutely right. I missed that ODatas `orderby` is missing in OpenSearch. So if we want filters we indeed have to include them locally.

I would prefer something along the lines of @valgur suggestion to provide a convenience function to return (Geo)Pandas DataFrames or any other format that can helps the users sort the results. But also not make pandas (or whatever convenience format we decide on) not a hard dependency for `sentinelsat`. A search&download tool should not depend on users being able to install numpy etc.

I'm still not a big fan of writing all the ordering functions into the code as this creates a lot of potential issues further down the road (see #65).

Fernerkundung:
Opensearch aka. Apache Solr *does* support sorting of results with a [sort parameter](https://wiki.apache.org/solr/CommonQueryParameters#sort).

So in theory this query should order by cloudcover in ascending order - however I can't get it to work. So either sorting is disabled or this field is not indexed.

```
https://scihub.copernicus.eu/apihub/search?q=%27(beginPosition:[2016-10-01T00:00:00Z%20TO%202016-12-11T00:00:00Z])%20AND%20(footprint:%22Intersects(POLYGON((-66.2695312%20-8.0592296,-66.2695312%200.7031074,-57.3046875%200.7031074,-57.3046875%20-8.0592296,-66.2695312%20-8.0592296)))%22)%20AND%20(platformname:Sentinel-2)%27&rows=100&sort=cloudcoverpercentage%20desc
```

lucadelu:
@Fernerkundung I agree with you that a new dependence (GeoPandas has a lot of other dependencies) should be avoid just for sorting and filtering, so what do you suggest? 

Also for me Apache Solr request is not working....

j08lue:
> Also for me Apache Solr request is not working....

Did the SciHub people say that when you asked (saw your Twitter chat)?

Fernerkundung:
@j08lue 

Their Twitter handler suggested I use EOSupport@copernicus.esa.int (the official support contact channel) and they created a ticket (CDS-19084). I will tell you the outcome as soon as they answer.

Fernerkundung:
I just got an answer to my ticket:

sorting by `cloudcoverpercentage` is not supported right now but will be in a future release.

I asked if there is a roadmap for said release and if any parameters allready support sorting but I guess in the meantime we have to roll our own solution.

---

Given that server side sorting will be supported in the future I'm okay with implementing simple client side sort functions that we can discard once this feature is enabled on SciHub. I'm also a fan of a convenience function to return DataFrames or something similar if a user explicitly wants to perform more advanced sorting. It is however quite tricky to turn a nested (Geo)JSON into a neat (Geo)DataFrame.

valgur:
I suggest creating a `.to_dict()` method that converts the response JSON to a list of dicts with the values in their appropriate Python types (e.g. `datetime` for timestamps). After that, converting this to a DataFrame should be as simple as passing that list of dicts to the DataFrame constructor. Creating a GeoDataFrame is probably easy as well, if you convert each WKT footprint to a Shapely geometry first.

If we have a `.to_dict()` implemented, then sorting that list will be straightforward even without using Pandas. We'll simply need to specify the dictionary keys to sort on when doing `sort()`.

Fernerkundung:
@valgur +1 on `.to_dict()`. That one would be handy even if `query` yields a generator (#65).

Fernerkundung:
Just heard back from the support team:

- server side sorting will be supported for all attributes except cloud cover
- a bug has been identified in the current version that prevented sorting
- new release is expected beginning 2017 which will allow sorting

So in other words we do need:
- [ ] an offline cloud cover sort
- [ ] a convenience function like `.to_dict()` for transferring the results to (Geo)DataFrames, databases, etc.

Any takers?

lucadelu:
@Fernerkundung I could work on this topic but I need more info how to solve the request.

- could the offline cloud cover sort be my proposal?
- what should be used in GeoDataFrames? we have to import the scenes as polygon?

Fernerkundung:
@lucadelu 

Sure you can propose an offline cloud cover sort.

I would not start with GeoDataframes. I'm +1 on @valgur suggestion of "a .to_dict() method that converts the response JSON to a list of dicts with the values in their appropriate Python types (e.g. datetime for timestamps)."`Anyway I think it is easiest if the two things are moved to two new, separate PRs - as this one has morphed into mainly discussion now.

Any input from @valgur, @j08lue, @willemarcel on this?

j08lue:
I agree with @Fernerkundung that it would be good to make new, separate PRs for sorting and dict output, respectively, and to avoid making Geopandas a dependency.

lucadelu:
I opened a new pull request for the to_dict function #96

lucadelu:
I updated the code and now should be following the comment, I added also tests

valgur:
Client-side product ordering is no longer necessary with server-side ordering support added in #123.

PR #71, opened 2016-11-24
some argument renaming upon request

j08lue:
As discussed in #63

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -122,7 +122,7 @@ class SentinelAPI(object):
         current value: 100 (maximum allowed on ApiHub)
     """
 
-    def __init__(self, user, password, api_url='https://scihub.copernicus.eu/apihub/', page_size=100):
+    def __init__(self, user, password, api_url='https://scihub.copernicus.eu/apihub/'):
         self.session = requests.Session()
         self.session.auth = (user, password)
         self.api_url = api_url
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -110,6 +110,9 @@ class SentinelAPI(object):
     api_url : string, optional
         URL of the DataHub
         defaults to 'https://scihub.copernicus.eu/apihub'
+    page_size : int
+        size of query pages on SciHub
+        default value: currently allowed maximum
 
     Attributes
     ----------
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -118,7 +118,8 @@ class SentinelAPI(object):
     api_url : str
         URL to the DataHub
     page_size : int
-        size of query pages on SciHub
+        number of results per query page
+        current value: 100 (maximum allowed on ApiHub)
     """
 
     def __init__(self, user, password, api_url='https://scihub.copernicus.eu/apihub/', page_size=100):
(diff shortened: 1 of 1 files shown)


valgur:
Very nice, thanks!

sentinelsat/sentinel.py
```diff
@@ -110,6 +110,9 @@ class SentinelAPI(object):
     api_url : string, optional
         URL of the DataHub
         defaults to 'https://scihub.copernicus.eu/apihub'
+    page_size : int
+        size of query pages on SciHub
```
Fernerkundung: I'd say this is rather "number of results per query page" as "size" can be ambiguous.

sentinelsat/sentinel.py
```diff
@@ -110,6 +110,9 @@ class SentinelAPI(object):
...
+        default value: currently allowed maximum
```
Fernerkundung: we should explicitly state the value (100) implemented - ESAs maximum limit might change in the future

Fernerkundung: we should explicitly state the value (100) implemented - ESAs maximum limit might change in the future

valgur: I agree. It should also include "optional"  after the type: `page_size : int, optional`.

valgur:
Looking at this parameter again, I'm starting to wonder if there is much point in making the `page_size` configurable (at least during initialization). I can't envision anyone actually reducing it from the maximum value. Given how unreliable the query responses from SciHub are, I would always try to get the response in as few queries as possible.

We could just initialize it to `100` in the constructor and if anyone really wishes to configure it then they can  do `api.page_size = 95`.

j08lue:
I agree (https://github.com/ibamacsr/sentinelsat/pull/71/commits/1bbd8eba9ef4cbbf5b399c5062a85a0f9370b72f).

j08lue:
now also with better docs following [your suggestions](https://github.com/ibamacsr/sentinelsat/pull/71#discussion_r89514033) @Fernerkundung.

sentinelsat/sentinel.py
```diff
@@ -117,9 +117,12 @@ class SentinelAPI(object):
         Session to connect to DataHub
     api_url : str
         URL to the DataHub
+    page_size : int
+        number of results per query page
+        current value: 100 (maximum allowed on ApiHub)
     """
 
-    def __init__(self, user, password, api_url='https://scihub.copernicus.eu/apihub/', max_rows=100):
+    def __init__(self, user, password, api_url='https://scihub.copernicus.eu/apihub/', page_size=100):
```
Fernerkundung: Setting the page_size on init is redundant due to L134 (setting page sized as fixed attribute).

Fernerkundung:
Looking good :+1: 

@valgur +1 on making the page size a class parameter. The Copernicus SciHub is the main portal for this tool. For every other national hub users can easily overrid it. If some national hubs (peps, australia, code-de,...) impose an even stricter limit we might need to expose this functionality in the CLI as well (in the future).

sentinelsat/sentinel.py
j08lue: right, missed that.

j08lue:
Ready to merge?

Fernerkundung:
I'm +1 on merging this. The merge itself has to be done by @willemarcel - who is the only one with write access to the repo.

PR #81, opened 2017-01-13
replaced self.products

lucadelu:
Hi,
In this pull request self.products is removed, now `query` is now returning a list of products, and the list is a parameter in the other functions using self.products

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -146,14 +146,11 @@ class SentinelAPI(object):
         query = self.format_query(area, initial_date, end_date, **keywords)
         return self.load_query(query)
 
-    def load_query(self, query, start_row=0, prods=None):
+    def load_query(self, query, start_row=0, prods=[]):
         """Do a full-text query on the SciHub API using the format specified in
            https://scihub.copernicus.eu/twiki/do/view/SciHubUserGuide/3FullTextSearch
         """
-        if not prods:
-            output = []
-        else:
-            output = prods
+        output = prods
         # store last query (for testing)
         self.last_query = query
 
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -146,11 +146,11 @@ class SentinelAPI(object):
         query = self.format_query(area, initial_date, end_date, **keywords)
         return self.load_query(query)
 
-    def load_query(self, query, start_row=0, prods=[]):
+    def load_query(self, query, start_row=0, products=[]):
         """Do a full-text query on the SciHub API using the format specified in
            https://scihub.copernicus.eu/twiki/do/view/SciHubUserGuide/3FullTextSearch
         """
-        output = prods
+        output = products
         # store last query (for testing)
         self.last_query = query
 
@@ -191,7 +191,7 @@ class SentinelAPI(object):
 
         # repeat query until all results have been loaded
         if total_results > self.page_size + start_row - 1:
-            self.load_query(query, start_row=(start_row + self.page_size), prods=output)
+            self.load_query(query, start_row=(start_row + self.page_size), products=output)
         return output
 
     @staticmethod
@@ -214,10 +214,10 @@ class SentinelAPI(object):
         query = ''.join([acquisition_date, query_area, filters])
         return query
 
-    def get_products_size(self,  prods):
+    def get_products_size(self,  products):
         """Return the total filesize in GB of all products in the query"""
         size_total = 0
-        for product in prods:
+        for product in products:
             size_product = next(x for x in product["str"] if x["name"] == "size")["content"]
             size_value = float(size_product.split(" ")[0])
             size_unit = str(size_product.split(" ")[1])
@@ -228,12 +228,12 @@ class SentinelAPI(object):
             size_total += size_value
         return round(size_total, 2)
 
-    def get_footprints(self, prods):
+    def get_footprints(self, products):
         """Return the footprints of the resulting scenes in GeoJSON format"""
         id = 0
         feature_list = []
 
-        for scene in prods:
+        for scene in products:
             id += 1
             # parse the polygon
             coord_list = next(
@@ -395,7 +395,7 @@ class SentinelAPI(object):
                 raise InvalidChecksumError('File corrupt: checksums do not match')
         return path, product_info
 
-    def download_all(self, prods, directory_path='.', max_attempts=10, checksum=False, check_existing=False, **kwargs):
+    def download_all(self, products, directory_path='.', max_attempts=10, checksum=False, check_existing=False, **kwargs):
         """Download all products returned in query().
 
         File names on the server are used for the downloaded files, e.g.
@@ -406,7 +406,7 @@ class SentinelAPI(object):
 
         Parameters
         ----------
-        prods : list
+        products : list
             List of products returned with self.query()
         directory_path : string
             Directory where the downloaded files will be downloaded
@@ -424,8 +424,8 @@ class SentinelAPI(object):
             (returned by get_product_info()). Product info is set to None if downloading the product failed.
         """
         result = {}
-        print("Will download %d products" % len(prods))
-        for i, product in enumerate(prods):
+        print("Will download %d products" % len(products))
+        for i, product in enumerate(products):
             path = join(directory_path, product['title'] + '.zip')
             product_info = None
             download_successful = False
@@ -444,7 +444,7 @@ class SentinelAPI(object):
                     traceback.print_exc()
                 remaining_attempts -= 1
             result[path] = product_info
-            print("{}/{} products downloaded".format(i + 1, len(prods)))
+            print("{}/{} products downloaded".format(i + 1, len(products)))
         return result
 
     @staticmethod
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -192,7 +192,6 @@ class SentinelAPI(object):
         # repeat query until all results have been loaded
         if total_results > self.page_size + start_row - 1:
             output += self.load_query(query, start_row=(start_row + self.page_size))
-            return output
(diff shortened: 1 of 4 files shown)


Fernerkundung:
I'm +1 on renaming the argument `prods` to `products`. The tests still need to be adapted to this API change. Apart from that this is looking good to go.

lucadelu:
@Fernerkundung `prods` renamed, I also updated README, cli.py and test_mod.py. Some tests are still failing, could you please check them?

j08lue:
The CI failures are all due to the query finding different products than we predicted, see e.g. https://travis-ci.org/ibamacsr/sentinelsat/jobs/192272911#L289 . But why is that? It seems to be an issue with the current `master` branch as well when I run the tests locally. Did SciHub for some reason start to return different results for the same queries?

j08lue:
> Did SciHub for some reason start to return different results for the same queries?

D'oh, that's probably the new single-tile products. I will revise the test queries and their respected results.

lucadelu:
@j08lue I cannot confirm your tests with master, for me they are working all fine and also for [travis the tests are passed](https://travis-ci.org/ibamacsr/sentinelsat/branches)

```
tests/test_cli.py::test_cli PASSED
tests/test_cli.py::test_returned_filesize PASSED
tests/test_cli.py::test_cloud_flag_url PASSED
tests/test_cli.py::test_sentinel1_flag PASSED
tests/test_cli.py::test_sentinel2_flag PASSED
tests/test_mod.py::test_format_date PASSED
tests/test_mod.py::test_convert_timestamp PASSED
tests/test_mod.py::test_md5_comparison PASSED
tests/test_mod.py::test_SentinelAPI_connection PASSED
tests/test_mod.py::test_SentinelAPI_wrong_credentials PASSED
tests/test_mod.py::test_api_query_format PASSED
tests/test_mod.py::test_invalid_query PASSED
tests/test_mod.py::test_format_url PASSED
tests/test_mod.py::test_format_url_custom_api_url PASSED
tests/test_mod.py::test_small_query PASSED
tests/test_mod.py::test_large_query PASSED
tests/test_mod.py::test_get_coordinates PASSED
tests/test_mod.py::test_get_product_info PASSED
tests/test_mod.py::test_get_product_info_scihub_down PASSED
tests/test_mod.py::test_get_products_invalid_json PASSED
tests/test_mod.py::test_footprints_s1 PASSED
tests/test_mod.py::test_footprints_s2 PASSED
tests/test_mod.py::test_s2_cloudcover PASSED
tests/test_mod.py::test_get_products_size PASSED
```

So, something is wrong with my implementation but I don't understand where

sentinelsat/sentinel.py
```diff
@@ -145,13 +144,13 @@ def query(self, area, initial_date=None, end_date=datetime.now(), **keywords):
...
-        self.load_query(query)
-        return self.products
+        return self.load_query(query)
 
-    def load_query(self, query, start_row=0):
+    def load_query(self, query, start_row=0, products=[]):
         """Do a full-text query on the SciHub API using the format specified in
            https://scihub.copernicus.eu/twiki/do/view/SciHubUserGuide/3FullTextSearch
         """
+        output = products
```
j08lue: @lucadelu This is the reason why the tests are failing. You switch back and forth between the variables `products` and `outputs`. This is actually not necessary. Remove `outputs` and just keep appending to `products`. This requires that you call the extraction from`json_feed` something else in [line 171](https://github.com/ibamacsr/sentinelsat/pull/81/files#diff-b41f2cfc13d00858923dd3e8da963187L171).

j08lue:
Right, I was too quick there drawing conclusions.

But I found the real reason. Please see my review comment https://github.com/ibamacsr/sentinelsat/pull/81/files#r97038859 .

sentinelsat/sentinel.py
lucadelu: I tried the following diff, but the failures are the same

```
diff --git a/sentinelsat/sentinel.py b/sentinelsat/sentinel.py
index b15f263..baa83fa 100644
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -150,7 +150,6 @@ class SentinelAPI(object):
         """Do a full-text query on the SciHub API using the format specified in
            https://scihub.copernicus.eu/twiki/do/view/SciHubUserGuide/3FullTextSearch
         """
-        output = products
         # store last query (for testing)
         self.last_query = query
 
@@ -168,14 +167,14 @@ class SentinelAPI(object):
         total_results = 0
         try:
             json_feed = content.json()['feed']
-            products = json_feed['entry']
+            entries = json_feed['entry']
             # this verification is necessary because if the query returns only
             # one product, self.products will be a dict not a list
-            if type(products) == dict:
-                products = [products]
+            if type(entries) == dict:
+                entries = [entries]
 
             # append to products
-            output += products
+            products += entries
 
             # get total number of returned results
             total_results = int(json_feed['opensearch:totalResults'])
@@ -191,8 +190,8 @@ class SentinelAPI(object):
 
         # repeat query until all results have been loaded
         if total_results > self.page_size + start_row - 1:
-            self.load_query(query, start_row=(start_row + self.page_size), products=output)
-        return output
+            self.load_query(query, start_row=(start_row + self.page_size), products=products)
+        return products
 
     @staticmethod
     def format_query(area, initial_date=None, end_date=datetime.now(), **keywords):
```

j08lue: Can you please push that, despite it not working, then it is easier to read. :-)

j08lue: I think it might be because your `products` list is not recursively updated when passed through the nested `self.load_query` calls. A NumPy array would. Try `products = self.load_query(...` in [line 194](https://github.com/ibamacsr/sentinelsat/pull/81/files/bef08af026834454d4118ffb9028920c89b72634#diff-b41f2cfc13d00858923dd3e8da963187R194).

j08lue: This might help: http://stackoverflow.com/questions/9896951/recursive-functions-and-lists-appending-extending

valgur:
```python
def load_query(self, ..., products=[]):
    output = products
    ...
    output += other_products
```
This will cause bugs. Python's default parameter values are shared across function calls, which is bad news for mutable objects like lists: http://docs.python-guide.org/en/latest/writing/gotchas/.

Consider doing
```python
output = []
output += products
```
instead to avoid this issue.

lucadelu: Ok, fixed in fc7fe1e
Now most of the tests pass, only two are still failing, `test_large_query` because  `assert len(products) > api.page_size` seems to be `100>100`; instead I'm not able to reproduce the error in `test_download_all`, if I follow the same code in a Python console it works.

lucadelu:
@Fernerkundung it is already like that, the problem seems to be that the `len(products)` is 100 and also the `page_size` is 100

sentinelsat/sentinel.py
```diff
@@ -145,13 +144,14 @@ def query(self, area, initial_date=None, end_date=datetime.now(), **keywords):
         and any other search keywords accepted by the SciHub API.
         """
         query = self.format_query(area, initial_date, end_date, **keywords)
-        self.load_query(query)
-        return self.products
+        return self.load_query(query)
 
-    def load_query(self, query, start_row=0):
+    def load_query(self, query, start_row=0, products=[]):
```
valgur: I suggest removing the `products=[]` argument altogether. All it is used for is to concatenate new products to an existing list. It makes a lot more sense to do 
```python
products = api.load_query(...)
products += api.load_query(...)
```
outside this method instead.

sentinelsat/sentinel.py
```diff
@@ -145,13 +144,14 @@ def query(self, area, initial_date=None, end_date=datetime.now(), **keywords):
...
         """Do a full-text query on the SciHub API using the format specified in
            https://scihub.copernicus.eu/twiki/do/view/SciHubUserGuide/3FullTextSearch
         """
+        output = []
```
valgur: Get rid of `output`.

Fernerkundung:
@lucadelu Then there's an error with the pagination. The number of returned products should be about 527 (give or take a few depending on reprocessing) - if it is exactly 100 that means that only the first page was queried, meaning that `output` probably contains only the first page with 100 products (the page size limit).

sentinelsat/sentinel.py
```diff
@@ -192,7 +192,8 @@ def load_query(self, query, start_row=0):
 
         # repeat query until all results have been loaded
         if total_results > self.page_size + start_row - 1:
-            self.load_query(query, start_row=(start_row + self.page_size))
+            self.load_query(query, start_row=(start_row + self.page_size), products=output)
```
valgur: Do `entries += self.load_query(...)` here and `return entries` on the next line.
This line is also the cause of the `test_large_query` failing, since the recursive call does not modify the `output` variable anymore.

lucadelu: Ok, now (with the last two commits) everything seems to work (one failing test was my fault)

sentinelsat/sentinel.py
```diff
@@ -192,9 +191,11 @@ def load_query(self, query, start_row=0, products=[]):
 
         # repeat query until all results have been loaded
         if total_results > self.page_size + start_row - 1:
-            self.load_query(query, start_row=(start_row + self.page_size), products=output)
+            output += self.load_query(query, start_row=(start_row + self.page_size))
+            return output
```
valgur: You can safely remove this line (`return output`). Other than that, everything looks good. :)

lucadelu: Done

Fernerkundung:
I'd say this is looking good to merge @willemarcel.

Thank you @lucadelu.

If everyone is fine with this I will adapt the docs in a separate PR to not create any more noise here.

PR #85, opened 2017-01-20
Remove api.products state

j08lue:


j08lue:
This is now all resolved in #81

PR #86, opened 2017-01-20
vcrpy improvements

valgur:
This PR adds command-line options to `py.test` that makes interacting with `vcrpy` recordings more comfortable. I also fixed the potential issue of response cookies being included in the recordings, which can leak session information.

The new `py.test` option is `--vcr` with the following choices:
* `use` – replays cassettes but does not record,
* `disable` – passes all queries directly to the server,
* `record_new` – replays cassettes and records any unmatched queries,
* `reset_all` – deletes all cassettes and re-records them.

So, if you do any adjustments to the code that affect unit tests, then you can simply add `--vcr record_new` or `--vcr reset_all` to `py.test` and can use `--vcr disable` to communicate directly with SciHub.


Fernerkundung:
Great addition! Should we worry about cassetes recording the wrong SciHub state - i.e. when `record_new` is passed while SciHub is down. I'm not sure if/how we could test for that.

valgur:
Yeah, it will record any incorrect responses right now. It would be useful to skip recording of failed tests but vcrpy does not support it currently: https://github.com/kevin1024/vcrpy/issues/205.
You can always use git to revert to a good recording or delete any bad new ones, though.

anujkumar93:
For anyone visiting this in the future: vcrpy doesn't provide an option in general to skip recording of failed tests. However, it can be done by simply deleting the cassette at the end of the test case if an exception exists in the stack trace. It's not exactly the same as skipping it but is pretty close to the desired behavior.

e.g. for someone using [vcrpy-unittest](https://github.com/agriffis/vcrpy-unittest/blob/master/vcr_unittest/testcase.py), you'd add a cleanup in the VCRMixin setUp:
```
        if self.vcr_enabled:
            self.addCleanup(self.cleanup_cassettes)
```,
where `cleanup_cassettes` is defined as 
```    
def cleanup_cassettes(self):
        if self.delete_on_exception:  # VCRMixin cls variable, should default to False for historical behavior
            if any(sys.exc_info()):
                vcr_path = os.path.join(self._get_cassette_library_dir(),
                                self._get_cassette_name())
                if os.path.isfile(vcr_path):
                    os.remove(vcr_path)
```

PR #87, opened 2017-01-20
Set the user agent header, reuse cookies across queries, improve homura's authentication

valgur:
I thought it would be nice towards the SciHub maintainers to include a user agent in queries. At the moment it is just "requests/*requests version*". Downloads do not specify a user agent since they use homura.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 requests
 click
-homura>=0.1.2
+homura>=0.1.4
 html2text
 geojson
 tqdm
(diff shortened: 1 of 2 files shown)


valgur:
This PR also changes the way `homura` authenticates with SciHub. Currently it uses session cookies for authentication, but this requires that at least one query is made using `self.session` first. However, there's a hiccup in `homura`, if `requests` stores cookies with identical names from multiple domains, which does happen in `sentinelsat` after the modifications in this PR. Because of this I [added a basic HTTP auth support to homura](https://github.com/shichao-an/homura/pull/14), which resolves this problem.

Issue #83, opened 2017-01-19
Use logging instead of print


j08lue:
I would like to use `sentinelsat` in a operational script where I need to log to files. I could redirect output from the CLI, but it would be much more convenient if I could use Python's `logging` system.

Could we replace all the `print` statements by logging?

Single challenge I am seeing: How does homura do the status IO and how can that be logged?

I would be ready to take on this task, if you support this.

valgur:
That would be nice to have, indeed.
Homura writes to `homura.STREAM` which defaults to `sys.stdout`. If it's important for you to also capture progress output from homura, then you could do something like

```python
import homura
import io

class HomuraIO(TextIOBase):
    def write(self, s):
        log.info(s)

homura.STREAM = HomuraIO()
```

It's also possible to set `show_progress=False` when initializing homura to just mute it.

j08lue:
Excellent, thanks @valgur. 

@willemarcel do you think you will accept #81? Then I would build on top of that to simplify merging.

Fernerkundung:
+1 on logging. I'd favour a clean cut where only the CLI prints something to stdout and everything done via Python uses logging - maybe with an added parameter to turn logging on/off.

Fernerkundung:
Can this be closed after #93 @j08lue?

Issue #91, opened 2017-01-25
Download improvement ideas


valgur:
I have been thinking for a while about possibly replacing `homura` with something custom-made based on `requests`. Here's my reasoning:
* A major drawback of homura is that we have no direct control over downloading, a significant part of sentinelsat.
* Homura does not check for HTTP error statuses, which can cause it to append junk to the downloaded files. This is a frequent situation on SciHub.
* This will open up the opportunity to add parallel download support. My idea here would be to split the download for each file into *n* parts, one for each download thread, each part being saved as "<*product*>.zip.part<*i*>", which will be merged together once the download is complete and checked. This naming scheme will make it easier to identify incomplete downloads in the download directory.
* [tqdm](https://github.com/tqdm/tqdm) can be used to show download progress, speed and ETA. An additional progressbar can also be added to display overall download progress for multiple files. tqdm is already used during checksum calculation in sentinelsat.
* Gets rid of the resuming of > 2 GB files [bug](https://github.com/ibamacsr/sentinelsat/blob/master/sentinelsat/sentinel.py#L388) in `pycurl`.
* If logging is implemented, then more fine-grained information and debug output can be logged.
* No need to worry about `certifi` being installed or not.
* By using `requests` the downloading functionality becomes mockable with `vcrpy` as well. We will be also able to simulate edge cases, like download errors, better for testing.

What are your thoughts on these proposed changes?

j08lue:
I do not have much experience with (the shortcomings of) `homura`, but I can see several advantages of getting rid of that dependency chain, one being that we/I/someone could easily make a QGIS `sentinelsat` plugin.

So the progress bar with `tqdm` (very nice choice) would be based on (parallel) iteration over download chunks? 

Ideally for me, the progress bar should be a lazy dependency and replaceable by a drop-in (e.g. by replacing an API class method).

The maximum number of download threads would be two, obviously?

Fernerkundung:
Great idea @valgur!

From what I can remember `homura` was implemented by @willemarcel mostly due to its convenience (integrated progress bar etc.) - so it is not set in stone as a dependency.

- removing a compiled dependency is a big, big plus, especially on Windows systems or for novice Python users (and enables easier integration into other tools like QGIS, etc.)
- we need to keep an eye on performance, but [curl vs. requests benchmarks](https://github.com/svanoort/python-client-benchmarks) look like large file transfers should be about the same speed, and small requests in the form of search is done already with `requests` anyway
[...]

PR #93, opened 2017-01-30
Logging

j08lue:
This is one of the many ways to implement logging (#83): SentinelAPI gets a `logger` attribute (logger name `sentinelsat.SentinelAPI`) and logs to that. The CLI also uses logging and sets a simple `StreamHandler` (logger name `sentinelsat`) that captures both, the messages from SentinelAPI and its own.

For the CLI, we could also use https://github.com/click-contrib/click-log, but that would add another dependency and not that much value (color messages).

The module does not add any handlers, following http://docs.python-guide.org/en/latest/writing/logging/
> It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers.
except when called through CLI.

Please let me know if you want something changed.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -32,6 +32,37 @@ Valid search query keywords can be found at the `ESA SciHub documentation
 <https://scihub.copernicus.eu/userguide/3FullTextSearch>`_.
 
 
+Logging
+-------
+
+Sentinelsat logs to ``sentinelsat`` and the API to ``sentinelsat.SentinelAPI``.
+
+There is no predefined `logging<https://docs.python.org/3/library/logging.html>`_ handler, 
+so in order to have your script print the log messages, either use ``logging.baseConfig``
+
+.. code-block:: python
+
+  import logging
+
+  logging.basicConfig(format='%(message)s', level='INFO')
+
+
+or add a custom handler for ``sentinelsat``
+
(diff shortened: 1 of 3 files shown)


j08lue:
NB: I did not implement logging of download progress (see https://github.com/ibamacsr/sentinelsat/issues/83#issuecomment-273779712) for now, since we are planning bigger changes to that part #91.

docs/api.rst
```diff
@@ -32,6 +32,37 @@ Valid search query keywords can be found at the `ESA SciHub documentation
...
+so in order to have your script print the log messages, either use ``logging.baseConfig``
+
+.. code-block:: python
+
+  import logging
+
+  logging.basicConfig(format='%(message)s', level='INFO')
+
+
+or add a custom handler for ``sentinelsat``
```
Fernerkundung: please add a reference to the implementation in `cli.py` - might help others in understanding the logger in action.

j08lue: Good idea. That would be a link to the code block on GitHub, right?

Fernerkundung: That would work, but I was thinking more along the lines of "as implemented in `cli.py`".

j08lue: That sounds better. I was hesitating to start creating links between docs and repo...

Issue #92, opened 2017-01-28
Keep a changelog


Fernerkundung:
With API breaking changes (as introduced in v0.8) I think it is high time we create a changelog. I propose one in the style of [keepachangelog.com](http://keepachangelog.com/en/0.3.0/).

I started working on one here: [942d920](https://github.com/Fernerkundung/sentinelsat/commit/a932f7674c9ba643fbc74e4aad7405aece570d97). I'm currently trying to propagate it based on historic PRs and commit messages - so the first iteration will need some review.

New PRs should add their changes into an 'Unrealeased' section, which allows maintainers to just change the version number w/o having to writing the changelog - anyone creating a PR probably knows best what they changed/broke/fixed.

Proposed layout is:

### Added
- for new features.
### Changed
- for changes in existing functionality.
### Deprecated
- for once-stable features removed in upcoming releases.
### Removed
- for deprecated features removed in this release.
### Fixed
- for any bug fixes.
### Security
- to invite users to upgrade in case of vulnerabilities.

lucadelu:
I like the proposal to add a CHANGELOG file it is really useful

do you want also to cover previous release or only from the next ones?

Fernerkundung:
@lucadelu I am trying to include previous releases as well - but it's quite trickly to puzzle everything together from PRs and commit messages.

Fernerkundung:
Closed with release of v0.8.1

PR #95, opened 2017-01-31
Changelog

Fernerkundung:
This PR is the first iteration of a changelog, as discussed in #92. I tried to populate old releases best I could from PRs and commit messages.

@j08lue and @valgur could you please look over the more recent versions (to which you contributed most) and correct me on mistakes.

Upcoming PRs should aim to populate the changelog as described in #92. In the future we should also describe this in a contribution guideline.


willemarcel:
Great! I thought about the necessity of a changelog some time ago, but I hadn't time to write it! Thank you very much!

PR #96, opened 2017-01-31
to_dict function

lucadelu:
I added function to return the products in dictionary type

PR #97, opened 2017-02-20
.to_dict() improvements, etc.

valgur:
I took the @lucadelu's .to_dict() pull request #96 and modified it a bit: added support for "double" data type, made use of `six` for Python version compatibility, changed the `dict` to `OrderedDict` to preserve the original order and made it a bit more compact overall.

Besides that I also made quite a few other changes that I did not feel like splitting into separate PRs (I'll split anything questionable to a separate PR, if necessary):
-  Added ``to_dict``, ``to_dataframe()`` and ``to_geodataframe()`` which convert the response content to respective types. The pandas, geopandas and shapely dependencies are not installed by default.
-  Renamed ``get_footprint()`` to ``to_geojson()`` and ``get_product_info()`` to ``get_product_odata()``.
-  Footprint output now makes use of `.to_dict()` and includes all of the properties included in a response (except for the footprint strings, which would be redundant).
-  ``KeyError('No results returned.')`` is no longer returned for zero returned products in a response. Empty result seems like a perfectly valid return value to me, no need to raise an exception. The CLI code also prints the number of returned products, so it's not required there either.
- Reorganized the package content somewhat: added underscores to functions and variables that are not intended to be used outside the package and moved them to the bottom of the class and file.
-  Got rid of unnecessary `url` and `content` instance variables.
- Fixed vcrpy cassettes recording login credentials in a request header by default (oooops....).


valgur:
I also removed Python 3.3 tests from TravisCI because `geopandas` depends on `matplotlib`, which no longer support Python 3.3 (https://travis-ci.org/ibamacsr/sentinelsat/jobs/203586985).
I could also remove the `pandas` and `geopandas` dependencies for tests and make these tests autoskip but I'd prefer running all the tests. Python 3.3 is over 4 years old by now.

Fernerkundung:
Great additions all around @lucadelu and @valgur. I suggest I'll handle the documentation update in a separate PR to not convolute this further.

@valgur wrt Python 3.3: My opinion is we should always strive to support the latest stable Python version (i.e. 3.6) and strive  for backwards compatability (2.7, <3.6) only when it doesn't hinder development too much. I'll add a section to the docs which states, that `sentinelsat` will still work with 3.3, just the `to_geopandas()` function is missing due to dependency conflicts.

Issue #88, opened 2017-01-23
Question about md5 checksum


melchisedech:
Hi all,
I have a problem with checksum control with sentinel sat.
Download is ok, but at the end of the process, there is an error with checksum control. I can't write the corrupt_scenes.txt:

Traceback (most recent call last):
  File "/home/geouser/anaconda2/bin/sentinel", line 11, in <module>
    sys.exit(cli())
  File "/home/geouser/anaconda2/lib/python2.7/site-packages/click/core.py", line 716, in __c                                   all__
    return self.main(*args, **kwargs)
  File "/home/geouser/anaconda2/lib/python2.7/site-packages/click/core.py", line 696, in mai                                   n
    rv = self.invoke(ctx)
  File "/home/geouser/anaconda2/lib/python2.7/site-packages/click/core.py", line 1060, in in                                   voke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/geouser/anaconda2/lib/python2.7/site-packages/click/core.py", line 889, in inv                                   oke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/geouser/anaconda2/lib/python2.7/site-packages/click/core.py", line 534, in inv                                   oke
    return callback(*args, **kwargs)
  File "/home/geouser/anaconda2/lib/python2.7/site-packages/sentinelsat/scripts/cli.py", lin                                   e 95, in search
    with open(os.path.join(path, "corrupt_scenes.txt"), "w") as outfile:
IOError: [Errno 20] Not a directory: u'/home/geouser/SENTINELSAT/DWNLD/S2A_OPER_PRD_MSIL1C_P                                   DMC_20160928T010718_R044_V20160927T233112_20160927T233110.zip/corrupt_scenes.txt'
geouser@ubuntu:~/anaconda2/bin$

Anyone have an idea ?
( i am sorry if it's not the correct canal for this question)

j08lue:
It seems like your `--path` argument points to a `.zip` file. It should be a directory.

Try again with `--path /home/geouser/SENTINELSAT/DWNLD`.

melchisedech:
Thank you for your help j08lue,
 i should have start with my command line (user and password have been changed):

[...]

PR #102, opened 2017-03-20
Add function to download_all pandas

j08lue:
So we have a nice `to_dataframe` function since https://github.com/ibamacsr/sentinelsat/pull/97.

But once you used the dataframe to filter or group the products returned by the query, how do you **download** them?

You could loop through their IDs with `api.download`. But wouldn't it be more convenient to be able to somehow pass that dataframe to `download_all`?

It turns out all `download_all` needs from the product dict are `id` and `title`. So here is a function that converts a dataframe to a list of dictionaries with these keys.

Now you decide @willemarcel if this is going too far with `@staticmethods` popping up everywhere... I think this is a logical addition to the `to_dataframe` and `to_geodataframe` methods.

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -215,6 +215,14 @@ class SentinelAPI(object):
         df.drop(['footprint', 'gmlfootprint'], axis=1, inplace=True)
         return gpd.GeoDataFrame(df, crs=crs, geometry=geometry)
 
+    @staticmethod
+    def dataframe_to_products_list(products_df):
+        """Convert dataframe created with to_dataframe back to list of products for download_all"""
+        products = []
+        for title, row in products_df.iterrows():
+            products.append(dict(title=title, id=row['id']))
+        return products
+
     def get_product_odata(self, id):
         """Access SciHub OData API to get info about a Product. Returns a dict
         containing the id, title, size, md5sum, date, footprint and download url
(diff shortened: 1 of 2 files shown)


Fernerkundung:
I don't really see a lot of added benefit, especially since this is specific to `to_dataframe` function - what is the logic that there are no download functions for `to_dict()`? What is the added benefit, since `download_all` also just loops through this list and uses `download` for each entry.

Additionally we already have the [documentation in place how to do this](https://sentinelsat.readthedocs.io/en/master/api.html#sorting-filtering), and I'd argue it doesnt get more straightforward/pythonic than this.:

```python
# download sorted and reduced products in order
for product_id in products_df_sorted["id"]:
  api.download(product_id)
```

valgur:
That's a good point you bring up @j08lue. This is indeed a gap in the code and the docs right now. Perhaps it is sufficient if we a note in the docs that you need to use `df.to_dict(orient='index')` as the input to `.download_all()`?

Actually, as @Fernerkundung's post hints, we don't really need the `title` parameter all that much at all. It is only used for returning some info about the download results. This functionality is of questionable quality anyway and in need of a rewrite (see #88). So, we can also consider modifying the `.download_all()` to simply take an iterable of product `id`-s as input instead of the above approach.

j08lue:
Good, thanks. I was actually having similar feelings towards this. It adds to the arbitrariness...

So should we not just remove `download_all`? I guess it is actually a relic of the stateful `SentinelAPI`.

valgur:
Well, the `download_all()` is a bit more robust than just a simple for loop: it retries the download in case of failures and prints some useful information as well. Besides, we need this functionality to be somewhere anyway for the CLI, might as well have it in a nice distinct function. I would definitely not consider removing it.

j08lue:
OK, I will just close this then. We can treat improving `download_all` in another issue.

Issue #104, opened 2017-03-28
How can the previous functionality of get_footprints() be replicated?


lpd2:
I have tried the to_geojson() method but this does not appear to save the footprint geometry. 

Thanks

valgur:
This is a definitely a bug. I submitted a fix in the #105 pull request.
Thanks for mentioning this!

Fernerkundung:
Fixed in #105.

PR #105, opened 2017-03-28
Fix invalid GeoJSON geometry output

valgur:
Reported in #104.

The GeoJSON Polygon features' geometries must be in the format of a list of list of coordinates. In my refactorings I accidentally changed it to just a list of coordinates, which produces invalid output. Fixed this.


valgur:
Great idea. Added that check to the existing `get_fooprints` tests.

Issue #101, opened 2017-03-20
Improve SentinelAPI.query interface


j08lue:
When using `SentinelAPI` from a scripts, I found it a bit tricky to figure out what to provide to the `SentinelAPI.query` function:

1. `area` - this needs to be a string of coordinates formatted in a special way. We have a function for reading a geojson file ([`get_coordinates`](https://github.com/ibamacsr/sentinelsat/blob/v0.9.1/sentinelsat/sentinel.py#L425)). But it includes both, **reading** the geojson file and **formatting** the string. I suggest to separate the two steps, so users can use the formatting function on its own with a coordinates sequence e.g. from `fiona` or some other source of geojson-like information.
2. `initial_date` and `end_date` - we parse these with [`_format_date`](https://github.com/ibamacsr/sentinelsat/blob/v0.9.1/sentinelsat/sentinel.py#L472) but that function **does not fail** when receiving some invalid input but just passes it on. I think fail it should.

The error the user gets from mistakes in the formats is some very very cryptic `SentinelAPIError` with lots Java traceback information from COAHub, because the query payload *is* generated but invalid. Very hard to figure out what went wrong.

Regarding the `area` parameter, we could also consider to do the string formatting **inside** the `query` function. Then the user just has to provide a (well-defined) geojson `coordinates` sequence. Would you support this, @willemarcel, @Fernerkundung and @valgur?

j08lue:
I see now that there was a **meaning** with `_format_date` passing values: The Open Access Hub supports things like `NOW` and `NOW-1DAY`. Some of these are listed in the [Open Access Hub query examples](https://scihub.copernicus.eu/twiki/do/view/SciHubUserGuide/5APIsAndBatchScripting#Open_Search_Queries_examples).

What should we do? Should we 

1. continue to pass (potentially invalid) values on
2. somehow validate non-date strings (e.g. `in_date.startswith('NOW')`) (and document which we allow)
3. or just not allow these special strings in our API, but only date strings and python dates?

j08lue:
Here is an implementation of the above suggestions:

1. Separate `format_coordinates_string` function for `area` parameter
2. Non-date strings are supported that fulfil `.startswith(`NOW`)`.

https://github.com/ibamacsr/sentinelsat/compare/master...DHI-GRAS:master

valgur:
These relative date values (`NOW`, etc.) seem easy enough to validate with a regex, so I support doing that. 

[...]

Issue #54, opened 2016-10-21
Sentinel-3 support


Fernerkundung:
This issue is supposed to be a starting point of discussion for Sentinel-3 support. As of 2016-10-20 [Sentinel-3 images are published on Scihub](https://scihub.copernicus.eu/news/News00109).
- [x]  include `--sentinel3` flag in CLI?
- [x] expose selection of [all Sentinel-3 instruments](https://sentinel.esa.int/web/sentinel/missions/sentinel-3/instrument-payload) in CLI/API or focus on OLCI?
- [x] does Sentinel-3 come with new search keywords? (cloud cover, etc.)
- [x] Do search&download work or need to be refactored?
- [x] Include tests once Sentinel-3 products leave pre-ops phase and are published on `apihub`

Fernerkundung:
@valgur @willemarcel @j08lue 

Do you think a `--sentinel3` flag would be usefull? This would return results from the OLCI, SLSTR and SRA instruments. Most users need only one of those products, which could be dealt with `-q 'instrumentshortname=SLSTR'`.

This becomes a more general question: Do we try to expose all instruments/products with flags in the CLI or do we expand the documentation with examples so users can use the `-q` flag to construct the query?

I got most of the Sentinel-3 support working with a simple change to the [`to_dict` parsing](https://github.com/Fernerkundung/sentinelsat/commit/ba3139e5c298a4fc2dd762c85177a3f9cf517353) due to a new item identifying collections as `{'name: 'collection, 'str':'002}` instead of ``{'name: 'collection, 'content':'002}``

valgur:
It would make sense to me to extend the CLI with `--sentinel3` and `--instrument=XXX` parameters. We should obviously strive to cover the most common query parameters (the instrument, location, date and cloud cover percentage, as far as I'm aware) in the CLI for less typing as well as for the convenience of not having to look up the relevant parameters in the SciHub API docs. Creating a separate parameter for each instrument seems excessive for me though. We can just list all of the relevant ones in the `--instrument` parameter info.

Thanks for fixing the `to_dict`, good job!

willemarcel:
I also agree with the `--sentinel` and `--instrument`!

Fernerkundung:
Good suggestions @valgur 

I propose to go forward as follows:

-[ ] deprecate `--sentinel1` and `--sentinel2` (with message) and remove them with the next major release
-[ ] add new `--sentinel <int>` flag that allows selection by `platformname`
[...]

PR #106, opened 2017-04-06
Make .to_dict() output the default, misc. other API improvements

valgur:
Comparing the `.to_dict()` output to the default `.query()` output, I have a hard time imagining anyone being interested in using the raw JSON response from the server. Because of this, I propose making the dictionary output the default for `.query()`, which this PR implements.

`.query()` now returns a parsed dictionary, while the `.load_query()` still returns the raw JSON and has been made a private method (prefixed with an underscore). To support full-text searches in a convenient manner, I added `.query_plain()` that takes just a query string and returns a parsed dictionary as well.

I also modified the dictionary structure slightly. I made the product ID is now the dictionary key instead of the product title. This is because of non-uniqueness of the product ID as the identifier. There are quite a few duplicate products with different IDs in the Data Hub (e.g. bit-by-bit identical products ending in '_0135' in [this query](https://scihub.copernicus.eu/apihub/search?q=%28beginPosition%3A%5B2014-01-01T00%3A00%3A00Z+TO+2014-12-31T00%3A00%3A00Z%5D%29+AND+%28footprint%3A%22Intersects%28POLYGON%28%28-66.2695312+-8.0592296%2C-66.2695312+0.7031074%2C-57.3046875+0.7031074%2C-57.3046875+-8.0592296%2C-66.2695312+-8.0592296%29%29%29%22%29+AND+%28producttype%3AGRD%29&rows=100&format=json)) and only one would be retained in the query output. Dropping duplicates would actually be very useful functionality as it would prevent unnecessary downloads, but I think this is not a good spot for doing that. We could consider adding an additional function for taking care of duplicates.

In addition to that major change I also updated the download code to expect a list of product IDs as input (so the parsed dictionaries work directly as well). This makes it easy to pass just a `DataFrame` index consisting of product IDs as the input to `.download_all()` as well. I also fixed the download error reporting bugs on the CLI interface reported in #88 and added unit tests to cover the CLI downloading code. This brings the unit test code coverage to near 100%. Only rare exception handling remains uncovered, which is perfectly fine.

I also added the improvement ideas in #101, although I seem to have missed @Fernerkundung's last comment.

There is still some work left to do regarding this PR. I have not updated the documentation yet, for one. I wanted to solicit your opinions first before doing that. Do you agree with the proposed changes?
I'm not sure about the method name for `query_plain`, for example. I think it could be named batter to reflect its purpose of doing (plain) full-text search.


valgur:
The first few commits are from #105, which has not been merged yet.

tests/test_mod.py
```diff
@@ -287,25 +325,30 @@ def test_footprints_s1():
...
-def test_footprints_s2():
-    api = SentinelAPI(**_api_auth)
-    products = api.query(
-        get_coordinates('tests/map.geojson'),
-        "20151219", "20151228", platformname="Sentinel-2"
-    )
+def test_footprints_s2(products):
+    footprints = SentinelAPI.to_geojson(products)
+    for footprint in footprints['features']:
+        print(footprint)
```
Fernerkundung: I don't think printing messages NOT from pytest is a good idea in this context - I think this might catch future contributors by surprise.

Fernerkundung:
Great contribution @valgur!
`query_plain` is just a more convenient shortcut to `_response_to_dict(_load_query(<query>))` but I think a good idea. I'd favour your intial naming convention of this as `query_raw` as the query is not expected in plain(text) but rather as a formated opensearch query.

Regarding the duplicate products with differing `UUID` - I don't think we should stress testing for this inside `sentinelsat`. This is a bug inside the SciHub server/database integration and thus should be dealt with on the server side. If we want to include this information I'd say it is best located in the Wiki/Docs, rather than as a further error handling mechanism.

The split of `get_coordinates` into the read and the formatting part I suggested in #101 can also be dealt with in a later PR. I still see it as a valid addition, because with this we could make the API and CLI 'geoJSON native' in that users can directly (re)use their JSON features they created with other tools such as rasterio, fiona or even other APIs.

👍 👍  for enabling `download_all(filtered_dataframe.index)`! A lot more straight forward than the current implementation.

tests/test_mod.py
valgur: Thanks for catching this. This was meant to be temporary, to help with debugging.

Fernerkundung:
@valgur Do you think this is ready to merge? If so I will build #54 on top of it to make merging easier for @willemarcel

valgur:
No, I would not. The latest commits are quite iffy. I had some ideas in mind about refactoring the downloading code, somewhat in preparation for #91, but I don't think it's a good approach anymore. I think I'll revert most of the later changes for now except for the extended OData querying functionality, which I think is very useful. Documentation needs updating, too. I'll see if I can fix that today.

valgur:
I modified the last commits and kept only the changes I'm happy with. I also documented my changes in the changelog and updated the docs in general.
That added [changelog entry](https://github.com/valgur/sentinelsat/blob/api-improvements/CHANGELOG.rst#010--2017-xx-xx) is also a good summary of the changes introduced by this PR.

The pull request should be ready for merging now. Feel free to let me know if you don't quite agree with anything, though.

valgur:
I find the (now quite powerful) metadata querying capability of `sentinelsat` to be a significant and useful feature which differentiates us from other similar utilities, so I suggest updating our tagline in the README to reflect that:

> Sentinelsat makes searching, retrieving the metadata and downloading of [Sentinel](http://www.esa.int/Our_Activities/Observing_the_Earth/Copernicus/Overview4)
> satellite images from the [Copernicus Open Access Hub](https://scihub.copernicus.eu/) easy.

Do you agree?

Fernerkundung:
+1 on changing the tagline. The ability to fine tune queries and now also the local filtering/analysis of the metadata is something I haven't seen in another library.

Great changes overall @valgur!

j08lue:
You are putting a lot of very good work into this, @valgur.

Issue #67, opened 2016-11-21
HTTP Status 500 - connect timed out


lucadelu:
Hi,

I'm new user of sentinelsat and I often get "HTTP Status 500 - connect timed out" response when I try to search. Is the problem on my request? (maybe the geojson is too big I'm trying to get scenes for Italy) or is the problem server side?

Thanks 
Luca

```
sentinel search --sentinel2 -s 20160701 -e 20160801 -c 5 USER PASSWD italia_bbox.json 

Traceback (most recent call last):
  File "/usr/local/bin/sentinel", line 11, in <module>
    load_entry_point('sentinelsat==0.7.1', 'console_scripts', 'sentinel')()
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/sentinelsat-0.7.1-py2.7.egg/sentinelsat/scripts/cli.py", line 86, in search
    api.query(get_coordinates(geojson), start, end, **search_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/sentinelsat-0.7.1-py2.7.egg/sentinelsat/sentinel.py", line 143, in query
    self.load_query(query)
  File "/usr/local/lib/python2.7/dist-packages/sentinelsat-0.7.1-py2.7.egg/sentinelsat/sentinel.py", line 158, in load_query
    _check_scihub_response(content)
  File "/usr/local/lib/python2.7/dist-packages/sentinelsat-0.7.1-py2.7.egg/sentinelsat/sentinel.py", line 98, in _check_scihub_response
    raise api_error
sentinelsat.sentinel.SentinelAPIError: (HTTP status: 500, code: None) 
# HTTP Status 500 - connect timed out

* * *

**type** Exception report

**message** _connect timed out_

[...]

PR #107, opened 2017-04-09
Cleaner exception messages

valgur:
I noticed that the server responses for errors include `cause-message` headers such as
```
cause-message: 'IOException : Server returned HTTP response code: 400 for URL:
          http://localhost:30333//solr/dhus/select?q=xxx:yyy&wt=xslt&tr=opensearch_atom.xsl&dhusLongName=Sentinels+Scientific+Data+Hub&dhusServer=https%3A%2F%2Fscihub.copernicus.eu%2Fapihub%2F&originalQuery=xxx%3Ayyy&rows=100&start=0&sort=ingestiondate+desc'
```
This provides essentially the same information but in a lot more concise and less intimidating manner than the wall-of-text stacktraces (such as in #67) have so far, so I modified the SentinelAPIError to use that instead.

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -404,15 +404,13 @@ class SentinelAPIError(Exception):
     """Invalid responses from SciHub.
     """
 
-    def __init__(self, http_status=None, code=None, msg=None, response_body=None):
-        self.http_status = http_status
-        self.code = code
+    def __init__(self, msg=None, response=None):
         self.msg = msg
-        self.response_body = response_body
+        self.response = response
 
     def __str__(self):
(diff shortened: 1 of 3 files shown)


Fernerkundung:
Great catch! I haven't seen these cause-messages before - looks like they have ben [added in December 2016](https://github.com/SentinelDataHub/DataHubSystem/search?utf8=%E2%9C%93&q=cause-message&type=)

willemarcel:
2 tests are failing, I believe it's due to this PR: https://travis-ci.org/ibamacsr/sentinelsat/builds/237145381

Could someone check it please?

PR #111, opened 2017-05-30
Fix tests that got broken by merging

valgur:
Related to #107


valgur:
I merged this right away since this is a straightforward and trivial fix. I hope that's ok.

valgur:
The unit tests and coverage calculation are not starting anymore. It appears we need to update our TravisCI and Coveralls integrations due to the organization name change.

PR #112, opened 2017-05-30
Update organization name

valgur:
Updated the organization name in docs and setup.py.

This should also trigger the first build for TravisCI under the new organization.


valgur:
Should we also perhaps add a [webhook](http://docs.readthedocs.io/en/latest/webhooks.html) for ReadTheDocs so the latest master branch commit is always used for documentation? I think @Fernerkundung has been updating the docs manually so far?

Fernerkundung:
I changed the organization name in RTD - new merges in master will now trigger new doc builds via Travis.

PR #113, opened 2017-05-30
Replace Coveralls with Codecov

valgur:
I have a small suggestion: let's replace [Coveralls](https://coveralls.io/github/sentinelsat/sentinelsat) with [Codecov](https://codecov.io/gh/sentinelsat/sentinelsat). You can take a look at the Codecov page for [pandas](https://codecov.io/gh/pandas-dev/pandas) for a good example of what is offered.

I find the Codecov user interface a lot cleaner and more useful than Coverall's and I also like their [browser extension](https://github.com/codecov/browser-extension), which makes it easy to see coverage straight inside GitHub source code view.

Here's what the badge looks like:
[codecov](https://codecov.io/gh/sentinelsat/sentinelsat)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,13 +11,13 @@ branches:
 before_install:
 - pip install -U pip
 install:
-- pip install coveralls
+- pip install codecov
 - pip install -e .[test]
 script:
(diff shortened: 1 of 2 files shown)


Fernerkundung:
👍 from me, especially since changing back would just mean changing the webhook.

PR #116, opened 2017-05-31
DataHub namechange

Fernerkundung:
Changing the documentation and comments from `Sentinels Scientific Data Hub (SciHub)` to the new name `Copernicus Open Access Hub`.


_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-slate
```
valgur: Is this commit intentional? It's quite old and not on the PR's branch.

Fernerkundung: It is not, and it's not supposed to be in this branch. It was however reverted with the later commit d1584e2. Should I rebase to remove it and create a new PR so it doesnt turn up in the master git logs?

valgur: You should try rebasing the current branch on the most recent `origin/master` commit. I'd expect that to fix this.

valgur: The changes look good otherwise 👍

PR #117, opened 2017-05-31
Replace pycurl with requests

valgur:
Replaces `homura`+`pycurl` with `requests` and `tqdm`, as discussed in #91.

The nice consequences of this are the significantly simpler installation instructions (just `pip install sentinelsat`) and that we can run all our tests without interacting with the server.

I have not run any benchmarks yet to check if there is any difference in download speeds when either `pycurl` or `requests` is used. I don't have time for this right now but you can volunteer to do that, if you feel like it.


j08lue:
Note so self: There was the [idea of making the progress bar replaceable](https://github.com/sentinelsat/sentinelsat/issues/91#issuecomment-275703332), which I need for making a QGIS wrapper for `sentinelsat`. That will come in the context of that implementation.

valgur:
@j08lue Thanks for reminding me about that. A good approach would be, I think, to refactor the downloading and checksum validation code to a separate class, which you can then replace or extend as required. I'll address that change in another PR/issue.

j08lue:
I compared the download speeds under Windows 10 and Python 3.6.1 and most recent `pycurl`, `curl`, etc. from `conda-forge` between `v0.10` and this PR's version.

Method: simply taking turn between two identical CLI calls, eye-balling the speeds over a couple of minutes each time.

Result: There was **no difference in the speeds**. In my case, both were around 5 MB/s.

I don't know how you could make a more accurate comparison. Maybe one could have two parallel downloads on different machines on the same internet connection over a longer time, comparing how many products of the same selection each version gets downloaded. But there are so many factors to control...

There certainly are no extreme differences.

j08lue:
> I'll address that change in another PR/issue.

@valgur That sounds like a very good approach. But take it easy, man, you just submitted this huge chunk of work we are reviewing here, we can take on those other things later. ;-)

And those will be changes that do not affect the API, at least not its public methods. So we can postpone that to post-1.0.

Fernerkundung:
I performed a small speed test with the following matrix:

- product: b7f2ce34-50d1-4dab-b88d-9250dc46f07c
- endpoints: https://scihub.copernicus.ei/apihub, https://code-de.org/dhus
- OS: Windows 7 x64, Ubuntu 16.04

All using Python 3.6 and the newest curl/pycurl version available through conda-forge. I let each pass run 2 times to only capture the download speed and not caching or search behaviour by the endpoints.

The download times:

- apihub:

| Ubuntu, requests | Ubuntu, curl | Win, requests | Win, curl |
| --- | --- | --- | ---|
| 5:41 | 5:52 | 6:33 | 6:21 |

- code-de:

| Ubuntu, requests | Ubuntu, curl | Win, requests | Win, curl |
| --- | --- | --- | ---|
| 6:13 | 6:22 | 6:03 | 6:11 |

The speed differences are negligible and are likely due to factors outside of the curl vs. requests comparison. Great work @valgur! This is a big step toward `v1.0`.

valgur:
Looks good, thank you both!

Fernerkundung:
@valgur please include the changes into the changelog. Afterwards I can publish it as a new release.

valgur:
Sorry, should have remembered to do that. I updated it now. Everything should be ready for releasing.

Issue #119, opened 2017-06-02
Contributing guidelines


Fernerkundung:
I think it is time for us to discuss the creation of contributing guidelines. Benefit would be that it helps others who want to contribute to this project but also as a documentation for the rest of us. I don't think we need to go all out on this like [pandas](http://pandas.pydata.org/pandas-docs/stable/contributing.html) or [numpy](https://docs.scipy.org/doc/numpy-dev/dev/index.html) but rather keep it relatively short, like [rasterio](https://github.com/mapbox/rasterio/blob/master/CONTRIBUTING.rst).

What are things we should include? I'll start with:

- [x] design principles / code style guide
- [x] developing environment / How to create a PR for a fix or feature?
- [x] testing / writing tests (Travis and local)
- [x] documentation and changelog
- [x] release and packaging

valgur:
Contributing guidelines would be nice to have, sure. We could create a GitHub wiki page for it initially, so it's easier to collaborate on writing it, and if there's anything that needs discussing we could do it here or on Gitter.

j08lue:
I made a first draft based on Rasterio's guidelines. It is not meant to be complete at all yet, just a starting point: https://github.com/sentinelsat/sentinelsat/wiki/CONTRIBUTING---draft

Fernerkundung:
@j08lue I built upon your initial draft and tracked the changes in a [public gist: CONTRIBUTE.md](https://gist.github.com/Fernerkundung/20c7eb237a16164a111d7a25fe75984b).

Any further comments by @valgur or @willemarcel, especially regarding the design principles and code conventions?

willemarcel:
It's good for me!

valgur:
Looks very good! I appreciate the fact that you managed to condense our goals to just a couple of bullet points. 😮

Some thoughts:
- The wording of "sentinelsat is meant to be not more and not less than a Pythonic client..." sounds slightly off, although I agree with the idea. Perhaps "nothing more and nothing less"?
- I would add that all unit tests must be able to run without interacting with the server by default, e.g. by using vcrpy cassettes. This is pretty much enforced anyway, tough, since I removed SciHub credentials from TravisCI.

j08lue:
> "nothing more and nothing less"

That is actually what I meant. Feel free to choose a less ambiguous formulation, too, if you like.

> all unit tests must be able to run without interacting with the server by default

Yes.

Fernerkundung:
Or just?
[...]

PR #120, opened 2017-06-02
Add a method to check if the OpenSearch query is too long

valgur:
I noticed that SciHub has an effective query length limit of about 2700–3600 bytes (depending on the content) when I tried to use a too detailed WKT polygon.
I experimented a bit and determined that the query size is effectively limited by the formatting and length of the server's internal query. I added a method `api.check_query_length()` that is able to calculate whether a query is too long with fairly good accuracy.
I also made the `SentinelAPIError` error messages for such situations more informative (the server reports back no useful information in this case).
I set up the relevant unit test to fail if the server's limit should change at some point in the future.

Edit: turns out the effective query length limit is approximately 3893 characters, but any special characters are counted twice towards that.

### DIFF
--- a/tests/test_mod.py
+++ b/tests/test_mod.py
@@ -1,5 +1,4 @@
 import hashlib
-import logging
 import textwrap
 from datetime import date, datetime, timedelta
 from os import environ
@@ -9,7 +8,6 @@ import py.path
 import pytest
 import requests
 import requests_mock
-from six import StringIO
 
 from sentinelsat import InvalidChecksumError, SentinelAPI, SentinelAPIError, geojson_to_wkt, read_geojson
 from sentinelsat.sentinel import _format_query_date, _md5_compare, _parse_odata_timestamp, _parse_opensearch_response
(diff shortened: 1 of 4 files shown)


Fernerkundung:
Should we maybe open a ticket with `eosupport@copernicus.esa.int` and ask for the exact query length? 

The only limitation I can find in the documentation is:
> The polygon describing the geographical area can have a maximum of 200 points that must be within an area described by 10 degrees of latitude and 10 degrees of longitude.

A quick look through the [DHuS Open Source Release](https://github.com/SentinelDataHub/DataHubSystem) also didn't turn up any limits so I am guessing it is enforced server-side, possibly by Solr.

tests/test_mod.py
```diff
@@ -202,26 +202,25 @@ def test_large_query():
 @pytest.mark.scihub
```
Fernerkundung: This test will only fail if cassettes are disabled or newly recorded. Which also means that it won't fail after configuration changes at the Copernicus server side. For the sake of speed of the tests I would still keep the vcr decorator but maybe add a line to the test documentation or testing instructions at the Readme or #119 that `--vcr record_new` should be performed on a semi-regular basis to catch such gremlins.

valgur:
Nice, thanks for finding a reference for the 8k character limit. 👍
I don't think that asking the support would help much, since this seems to be a quite arbitrary limit which is not explicitly specified anywhere in the server-side configuration, I think.

By the way, this PR also adds a workaround for the minor issue of plus symbols getting misinterpreted as spaces on the server side. This is a clear bug in the DHuS, so I opened an issue: https://github.com/SentinelDataHub/DataHubSystem/issues/23

valgur:
And the 200-point polygon limit that is mentioned in the docs appears to be just a consequence of this query limit. When experimenting with the query limits I managed to get successful responses with polygons  containing more than 300 points. The only problem is that you have to reduce the number of decimal places significantly to be able to fit that many in a query.

Fernerkundung:
> containing more than 300 points. The only problem is that you have to reduce the number of decimal places significantly to be able to fit that many in a query.

I went up to 400 once. With the new function you added one can easily test by how much the decimal accuracy has to be reduced for the query to succeed, without brute forcing the Open Access Hub, which is an added bonus. 👍

PR #121, opened 2017-06-03
Add functionality to limit the number of returned products

valgur:
This PR adds `limit` and `offset` parameters to `query()` and `query_raw()` and `--limit` to the CLI.

It's useful to have a way to limit the number of returned results. You probably don't want to receive thousands of results when testing out queries or you might want to get just *n* most recent images for an area, for example.


valgur:
Two closely related improvement ideas I've been thinking about:
* Limit the number of returned items by default. To 1000 items, perhaps. It's not difficult to query a large portion of the whole SciHub by accident and we currently do not provide feedback on the number of found products before they are returned. We should print a warning when that limit is in effect, of course. An even lower limit of 100 might be even better to make it more obvious that there's a limit in place by default.
* Add a way to get the number of products matching a query on the server via the API. Something like `query(..., return_count=True)`, which would then return a count of matching products in addition to the products dictionary. The CLI should always print the number of found products to the log anyway.

What do you think guys?

j08lue:
Good ideas, keep them coming. :-)

> Limit the number of returned items by default.

I support this, if it is obvious for the user how to override it. 100 is a reasonable default limit, I find.

> Add a way to get the number of products matching a query on the server via the API

Is it not easy enough for the user to check the count with `len(results)`? I think it is best to avoid functions returning different things based on conditions.

Fernerkundung:
> Limit the number of returned items by default.

What is the advantage of this? If a user queries a large area over a long time this will return the first page of results (if the default is 100), but the query will still be executed on the DHuS. As long as the DHuS doesn't split up long queries into multiple smaller queries that produce about 100 results this will not considerably reduce server load. It is possible that such behaviour might be implemented in the future but the current DHuS release does not contain anything alike in its code (I might be mistaken though, my Java game is pretty weak).

I do support having a reasonable default, that limits the amount a new user queries. In my opinion we already have such a default in place with the default time queried being NOW-1DAY to NOW. If a user extends that timeframe he deliberately wants more results and I suspect most if not all will extend the limit to a large number then anyway.

> Add a way to get the number of products matching a query on the server via the API.

I'm with @j08lue on this. If it is as easy as `len(products)` I don't see an advantage of including it as functionality in the API.

j08lue:
> If a user queries a large area over a long time this will return the first page of results (if the default is 100), but the query will still be executed on the DHuS.

Right, if the full query is executed on the server anyways, this `limit` on `&rows` does not help.

valgur:
Good points, @Fernerkundung. You might be correct regarding the server load. It seems to me, though, that they run the query again for each of the 100-entry subqueries and don't do any caching, so I would assume that limiting the amount of subqueries would be beneficial for the server.

You make a good point that we already have a reasonable and effective limit in place by restricting the query to just the last 24 hours by default. I agree that this is quite effective limit by itself and I no longer think that adding hard limit to the response size would be a good idea. It would be a potential source of confusion and issues for our users without much benefit.

Regarding adding an option to also return the number of matched products, I see its main benefit in being able to easily get the available number of products for some query without needlessly stressing the server. The situation I have in mind here is something like "get the total number of products that cover country X".
Do you see this too much of an edge case to be included in the API?

Fernerkundung:
> I see its main benefit in being able to easily get the available number of products for some query without needlessly stressing the server

Are we able to get the number of results without querying all of them? That might indeed reduce server stress, depending on the way the database works internally.

But then your proposed solution would still query all results anyway, since it returns the full results in addition to the count:

> `query(..., return_count=True)`, which would then return a count of matching products in addition to the products dictionary. 

If such a counting method significantly reduces the server load then I'd say it would best be implemented in its own method, maybe `query_count` or something alike. 

We are however at a point where we are speculating if it reduces server load. How can we be sure it works as expected? Maybe we are at a point where we would need to ask one of the Open Hub engineers.

valgur:
> Are we able to get the number of results without querying all of them? That might indeed reduce server stress, depending on the way the database works internally.

The total number of products that matched the query is provided in the `opensearch:totalResults` field. `sentinelsat` currently discards this when the response is converted to a dictionary. 

> But then your proposed solution would still query all results anyway, since it returns the full results in addition to the count

I forgot to mention that I intended the `return_count` flag to be always used together with `limit` applied.

> We are however at a point where we are speculating if it reduces server load. How can we be sure it works as expected? Maybe we are at a point where we would need to ask one of the Open Hub engineers.

I don't think it is particularly relevant, really. I did a quick test with 56k products in a response and it took 13 minutes to complete. As a user I would not want to wait that long even if it does not affect the server significantly in any way.

> If such a counting method significantly reduces the server load then I'd say it would best be implemented in its own method, maybe query_count or something alike.

I have considered doing it this way. Making it a separate method would mean that we would either have to keep it consistent with `query`'s input params or keep it simple and only support raw query strings, though. Maybe the latter and call it just `count()` since `query_count()` sounds too much like "count of queries" to me? Though, I would still prefer making it part of `query()` above that solution.

valgur:
I added an implementation of `count()` and made the CLI log the number of products matched. Let me know if you support this change or not.

Fernerkundung:
I like the compromise the that parsing for `total_results` happens in `_load_query` so we can keep the expected return of `query` stable and have a `count` method that efficiently parses available data for large queries.

Great work again @valgur :+1:

PR #123, opened 2017-06-03
Add server-side order-by support

valgur:
This adds an `order_by` parameter to `query()` and `query_raw()` and `-o,--order-by` to the CLI.
The input to these parameters is just a comma-separated list of keywords based on which to sort.
A minus sign can be added as a prefix to a keyword to sort it in descending order. E.g. "cloudcoverpercentage, -beginposition".

It appears that the server-side `&orderby=...` support has been available for a while but it's format is not too obvious: you need to always include "asc" or "desc" along with the keywords, e.g. "cloudcoverpercentage asc, beginposition desc".


valgur:
I'll address the documentation and changelog updates in another PR. It's easier to update them all at once.

Issue #125, opened 2017-06-07
Duplicate product on SciHub causes sentinelsat download to freeze


justinkotzebue:
When I attempt to download S2A_MSIL1C_20170205T105221_N0204_R051_T31TCF_20170205T105426, which exists twice on SciHub

- https://scihub.copernicus.eu/dhus/odata/v1/Products('2f379a52-3041-4b92-a8a8-92bddc495594')/$value
- https://scihub.copernicus.eu/dhus/odata/v1/Products('10762fc0-461d-48d7-ad93-7ce71af9bd1a')/$value

the product (probably in one of the two versions) is downloaded correctly. But afterwards, the download freezes at 0%.

j08lue:
We notice quite often that SciHub publishes products multiple times, where the first release is faulty (data all NaN or missing files) and the second is a fixed version. Mostly with slightly different production time stamps, so there are no issues with downloading both.

In this case, however, the production times are identical. The downloader finds the existing file name and freezes.

This will be very easy to demonstrate once we have the `--title` option from https://github.com/sentinelsat/sentinelsat/issues/115. But I will make a test example too.

Fernerkundung:
@justinkotzebue In the case of two non-identical products with identical (file)names what would you expect as behaviour? An error message, both downloaded and renamed with appendixes, a kind of check if one of them is faulty?

I'm a bit stumped by this behaviour of the OpenAccessHub, since, in theory, products with the same sensing and production timestamps (hence the identical filenames) should be identical in content. Therefore it shouldn't matter which one `sentinelsat` downloads. Your example however clearly demonstrates that there are faulty products published under identical filenames.

j08lue:
Checking for faults would be nice, but that would be very heuristic (guessing at files that should be there, ...) or heavy (checking for all-NaN files). Checksums obviously do not help here, because the they are generated from the released products on server side as well as on client side and are therefore identical even for faulty releases.

It would be quite cool if `sentinelsat` could remove duplicates already in the OpenAccessHub results based on identical **granule(s)** and **sensing time** (i.e. ignoring the production time). Like with an option `--deduplicate-method latest`. Sure, users can do that after the download, but that means they already pulled a lot of unusable data. But would that become too automagic?

Fernerkundung:
> they already pulled a lot of unusable data

Any examples how often duplicate products with identical filenames actually appear? Until shown otherwise my feeling is this is more of an edge case where some kind of error message would suffice.

[...]

PR #126, opened 2017-06-08
Use a temporary file name for files being downloaded

valgur:
This PR appends '.incomplete' to files being download to make it more obvious that they are incomplete. If a file exists with the true filename, then skip downloading that file and assume that it is complete and verified.

This also fixed the problem described in #125 (tested it on my machine). The server already sorts the products by the most recent ingestion date anyway and the modified download function skips downloading for existing files with the same name, which means that there is actually no need for additional deduplication code. There are no warnings being printed in such situations, though.

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -373,14 +373,25 @@ class SentinelAPI:
         skip_download = False
         if exists(temp_path):
             if getsize(temp_path) > product_info['size']:
+                self.logger.warning(
+                    "Existing incomplete file {} is larger than the expected final size"
+                    " ({} vs {} bytes). Deleting it.".format(
+                    str(temp_path), getsize(temp_path), product_info['size']))
                 remove(temp_path)
             elif getsize(temp_path) == product_info['size']:
                 if _md5_compare(temp_path, product_info['md5']):
                     skip_download = True
                 else:
+                    # Log a warning since this should never happen
+                    self.logger.warning(
+                        "Existing incomplete file {} appears to be fully downloaded but "
+                        "its checksum is incorrect. Deleting it.".format(
+                            str(temp_path), getsize(temp_path), product_info['size']))
                     remove(temp_path)
             else:
                 # continue downloading
+                self.logger.info("Download will resume from existing incomplete file "
+                                 "{}.".format(temp_path))
                 pass
 
         if not skip_download:
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import, division, print_function
 import hashlib
 import logging
 import re
+import shutil
 import xml.etree.ElementTree as ET
 from collections import OrderedDict
 from contextlib import closing
@@ -326,7 +327,7 @@ class SentinelAPI:
         values = _parse_odata_response(response.json()['d'])
         return values
 
-    def download(self, id, directory_path='.', checksum=False, check_existing=False):
+    def download(self, id, directory_path='.', checksum=False):
         """Download a product.
 
         Uses the filename on the server for the downloaded file, e.g.
@@ -343,10 +344,6 @@ class SentinelAPI:
         checksum : bool, optional
             If True, verify the downloaded file's integrity by checking its MD5 checksum.
             Throws InvalidChecksumError if the checksum does not match.
-            Defaults to False.
-        check_existing : bool, optional
-            If True and a fully downloaded file with the same name exists on the disk,
-            verify its integrity using its MD5 checksum. Re-download in case of non-matching checksums.
             Defaults to False.
 
         Returns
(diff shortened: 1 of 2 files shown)


j08lue:
How about checksumming? Is that done on the `.incomplete` files or the complete ones? 

It might be good to be able to verify complete files, too.

j08lue:
This will not completely resolve #125 with identical file names, because sentinelsat might download the faulty product first and regard it as complete. But with [the explicit resolution of duplicates we were discussing](https://github.com/sentinelsat/sentinelsat/issues/125#issuecomment-307127024), we are fine.

sentinelsat/sentinel.py
```diff
@@ -318,28 +315,42 @@ def download(self, id, directory_path='.', checksum=False, check_existing=False)
...
+            # We assume that the product has been downloaded and is complete
+            return product_info
+
+        # Use a temporary file for downloading
+        temp_path = path + '.incomplete'
+
+        skip_download = False
+        if exists(temp_path):
+            if getsize(temp_path) > product_info['size']:
+                remove(temp_path)
```
j08lue: So a download finding an existing `.incomplete` files will not be continued? 

Did we already drop continuing partial downloads with the change to `requests`?

sentinelsat/sentinel.py
```diff
@@ -318,28 +315,42 @@ def download(self, id, directory_path='.', checksum=False, check_existing=False)
 
         self.logger.info('Downloading %s to %s' % (id, path))
 
-        # Check if the file exists and passes md5 test
-        # The download function will by default continue the download if the file exists but is incomplete
-        if exists(path) and getsize(path) == product_info['size']:
-            if not check_existing or _md5_compare(path, product_info['md5']):
-                self.logger.info('%s was already downloaded.' % path)
-                return product_info
+        if exists(path):
```
Fernerkundung: Why abandon the feature of `check_existing`? I find it very useful to test the integrity of files I am given or acquired through other means than the Open Access Hub. check_existing=True is a feature that allows me to quickly compare a folder of files to the MD5 checksums published at the DHuS.

It could be easily implemented at this point, with the current logic beeing moved to an `elif`.

Fernerkundung:
I like the `.incomplete` files. However, I am still in favour of an explicit deduplication with warning.

> server already sorts the products by the most recent ingestion date anyway and the modified download function skips downloading for existing files with the same name

What if a user uses `orderby`? Queries with duplicates might then chose either file depending on what you order by. This would be unexpected behaviour in my opinion.

j08lue:
> I am still in favour of an explicit deduplication with warning.

I agree. I see this as an additional feature to de-duplication.

sentinelsat/sentinel.py
valgur: Partial downloads are definitely continued, even with this change. `getsize(temp_path) > product_info['size']` situation deletes the temporary file because its size on disk is larger than it is supposed to be, which should never happen. I should probably log a warning in that situation, though.

j08lue: Oh, sorry, my misunderstanding.

sentinelsat/sentinel.py
valgur: I felt that it is a bit orthogonal to the main functionality of `download()`. Though, I agree, I have found this functionality useful quite often myself, too. Since the `.zip` files can now be expected to be complete and verified with this PR's changes, there is normally no need to recheck existing files.
There are a couple of options:
* Add a separate method for this (`checksum()`?) functionality, which takes a list of `id`s as input.
* Not implement this at all. The zip files can be verified quite easily with `zip -T` or `7z t` commands or via a GUI.
* Re-add it to `download()`. Since you are likely interested in fixing the broken files rather than just deleting them, `download()` is actually a quite sensible place for this functionality.

I guess `download()` is still the best place for this feature. :)

valgur:
I agree, that this only partially solves the issue and additional deduplication with explicit warnings would be preferred. The problem with adding that functionality is that right now `download_all()` assumes the input to be just a list of product IDs, so it would either have to make its input more strict or run `get_product_odata()` for each product first.
A reasonable solution might be to add a new method `remove_duplicates()` or something like that, which is called in the CLI by default before `download_all()`. On the API side it would be the responsibility of the user to call it before `download_all()` or to take care of deduplication themselves in whatever way.

sentinelsat/sentinel.py
Fernerkundung: I see your point in removing it. How about another option:

- set `checksum=True` as default und move it up into `if exists(path)`. Use it as a single option to check temp as well as completed files.

You already proposed in another PR that setting `checksum=True` is a sensible default - and I agree. We can use this option to cover current as well as past downloads. Only downside I see is a use case where someone periodically runs the same, large query with the goal to only download new products. Then however they could adapt the query or just use the `limit` keyword.

j08lue:
> A reasonable solution might be to add a new method remove_duplicates() or something like that, which is called in the CLI by default before download_all(). On the API side it would be the responsibility of the user to call it before download_all() or to take care of deduplication themselves in whatever way.

Yes, that would be a very transparent solution. I think it is reasonable behaviour, too, that `download_all` with a list of IDs actually gets all these products and does not filter them.

sentinelsat/sentinel.py
valgur: I certainly agree with the suggestion of making `checksum=True` default, but I don't support making the checking of existing files the default and only behavior. I use `sentinelsat` exactly as you described here: to run `download_all()` periodically with a full list of products against a folder with many existing files, expecting them to be skipped. This is useful if the server happened to be down during the last invocation, for example.
So, I suggest simply restoring the old behavior.

Fernerkundung: Agreed. After thinking about it I do support your proposition
> Add a separate method for this (checksum()?) functionality, which takes a list of ids as input.

Local checksum tests are not a `download` but rather a separate function, that I would expect to call with either a folder (using the filenames as reference), a list of product IDs or a full OData dictionary. I suspect this would then be a but much for this PR - so I'd say you leave out `check_existing` in this PR and we'll deal with it in another Issue/PR.

PR #132, opened 2017-06-09
Make the number of decimal figures configurable in geojson_to_wkt(), improve unit test fixtures

valgur:
This is a small follow-up to #120. Changing the number of decimal figures in WKT coordinates makes it easier to adjust the length of your query to fit the server's limit.
I also reduced the default number of decimal places to 4 digits. This corresponds to about 10 m rounding error, which should be more than enough considering the fact that the footprint polygons are not all that accurate anyway.


valgur:
The cassettes in unit tests were getting pretty large for a Git repo, so I created a custom serializer for the cassettes that stores and reads any binary data (that is, downloaded products) separately from the `.yaml` files. This reduces the cassettes' total size from 18 MB to just 3.5 MB.

I also moved all unit test fixtures (including cassettes) to a separate `tests/fixtures/` folder to make things a bit more organized.

sentinelsat/sentinel.py
```diff
@@ -127,16 +127,16 @@ def format_query(area=None, initial_date='NOW-1DAY', end_date='NOW', area_relati
 
         query_parts = []
         if initial_date is not None and end_date is not None:
-            query_parts += ['(beginPosition:[%s TO %s])' % (
+            query_parts += ['beginPosition:[%s TO %s]' % (
```
j08lue: Nice catch to save some more characters...

Issue #128, opened 2017-06-09
Method to checksum existing products


valgur:
#126 removes the `check_existing` option from `download()` and `download_all()`, as it is quite unrelated to the main purpose of these functions. The main use case for that feature was to check the integrity of already downloaded files, which is potentially quite useful.

There are a couple of options regarding this feature:
* Add a separate method for this (`checksum()`?) functionality. Its input could be a list of product ids and a directory or a list of file paths.
* Not implement this at all. The zip files can be verified quite easily with `zip -T` or `7z t` commands or via a GUI.
* Re-add it to `download()`. Since you are likely interested in fixing the broken files rather than just deleting them, `download()` is actually a quite sensible place for this functionality.

The consensus so far in #126 was to make it a separate function.

Fernerkundung:
From a design perspective I would expect such a function to work with the following use cases:
- call with a folder of downloaded files (using the filenames as reference)
- list of product IDs and corresponding folder of files or list of files
- full OData dictionary and corresponding folder of files or list of files

j08lue:
If this becomes a full-fledged function, maybe we should not collapse the CLI into one command (`sentinelsat`) anyways but keep it a group with sub-commands?

Fernerkundung:
I don't think every option the API offers has to be exposed to the CLI which is supposed to be a user friendly way to use the API. If we include every option it becomes too convoluted.

That being said, we still include the most common use cases, so I'd say still merge it and include a flag like `--checksum` that takes all query elements and checks the integrity of downloaded files. In turn I would remove the `--no-md5` option proposed in #129 and just use the default API behaviour for the CLI.

valgur:
Resolved by #133.

PR #133, opened 2017-06-09
Add check_existing() method

valgur:
Implements #128.

Supported inputs are a list of file paths or a list of product IDs and a directory.
Returns a list of corrupt or missing files.

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -564,17 +564,20 @@ class SentinelAPI:
 
         return output
 
-    def check_existing(self, paths=None, ids=None, directory=None, delete=False):
+    def check_files(self, paths=None, ids=None, directory=None, delete=False):
         """Verify the integrity of product files on disk.
-
-        The input can be a list of products to check or a list of IDs and a directory.
 
         Integrity is checked by comparing the size and checksum of the file with the respective
         values on the server.
 
+        The input can be a list of products to check or a list of IDs and a directory.
+
         In cases where multiple products with different IDs exist on the server for given product
         name, the file is considered to be correct if any of them matches the file size and
-        checksum.
+        checksum. A warning is logged in such situations.
+
+        The corrupt products' OData info is included in the return value to make it easier to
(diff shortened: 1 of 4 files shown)


sentinelsat/sentinel.py
```diff
@@ -527,6 +527,124 @@ def check_query_length(query):
...
+
+    def check_existing(self, paths=None, ids=None, directory=None, delete=False):
+        """Verify the integrity of product files on disk.
+
+        The input can be a list of products to check or a list of IDs and a directory.
+
+        Integrity is checked by comparing the size and checksum of the file with the respective
+        values on the server.
+
+        In cases where multiple products with different IDs exist on the server for given product
```
Fernerkundung: Should we raise a warning in these cases to inform users that there might be multiple or newer products available? Might be relevant if the existing file returns a correct checksum but contains invalid data and the later ingested file might be correct. (as reported in #125).

j08lue: Good idea. That could be in [this line](https://github.com/sentinelsat/sentinelsat/pull/133/files/8d215f0c3adf41425b9132b1984781e443b9c029#diff-b41f2cfc13d00858923dd3e8da963187R626) - checking whether `product_infos` for any `name` is longer than 1. It would not need to contain much information on the products but direct the user to use the (planned) de-duplication method.

valgur: I agree, let's log a warning in that case. I'll add INFO level log messages of corrupt files as well.

sentinelsat/sentinel.py
```diff
@@ -527,6 +527,136 @@ def check_query_length(query):
...
+
+        # Group the products
+        output = OrderedDict((name, dict()) for name in names)
+        for id, metadata in products.items():
+            name = metadata['identifier']
+            output[name][id] = metadata
+
+        return output
+
+    def check_files(self, paths=None, ids=None, directory=None, delete=False):
```
j08lue: I like the new name. It is more instructive.

PR #134, opened 2017-06-10
Add pypy to TravisCI build environments

valgur:
Just curious to see whether `sentinelsat` works on pypy as well. It should not affect our build time since all builds are run in parallel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
 - 3.5
 - 3.6
 - pypy
+- pypy3
 cache: pip
 branches:
   only:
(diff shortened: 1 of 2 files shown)


valgur:
The result is: no, it does not work. Pypy does not support numpy, so pandas will not work.

Fernerkundung:
> Pypy does not support numpy, so pandas will not run

Which is only a dependency for `to_(geo)dataframe` and no one would try to use those when they can't install `pandas` in the first place.

I'd say we can say it works on Pypy, bar this small convenience function.

j08lue:
I don't know whether you would like to spend more time on this. 

But if you do, could we mark that test somehow to only run when `pandas` is installed? It is a conditional import anyways.

Fernerkundung:
Great to know @valgur! That's something we can mention in the docs and point to this closed PR should someone be inclined to use `sentinelsat` with PyPy.

PR #136, opened 2017-06-15
Improve logging

valgur:
Add some new log messages to see query progress with DEBUG.
Format log messages only when they are being emitted.


j08lue:
You are unstoppable, @valgur ... Those are good improvements.

PR #137, opened 2017-06-16
Merge query() and raw_query()

valgur:
Since `query()` is really just an extended wrapper around `query_raw()`, we could easily further extend `query()` a bit and get rid of `query_raw()` entirely.

This PR adds parameter `raw=` to `query()` and removes the `query_raw()` method. The `raw` string is just appended to the main query string with an `AND` in between. So, to get the equivalent functionality of `query_raw()` you can do `query(None, None, None, raw="query")` or `query(raw="query", initial_date=None, end_date=None)` now. Which, I have to admit, is kind of ugly. We will also have to prohibit ourselves from adding any `query()` parameters that would change the query string by default (as the date params do right now).

My motivation stems from the way I have typically used `query_raw()` in my own code. That is:
```python
query = api.format_query(aoi, initial_date, end_date)
query += " AND some additional query parameters"
products = api.query_raw(query)
```

I also reordered some `query()` args so the ones that will be passed to `format_query()` are always the first. This allowed me to make the signature of `count()` identical to `query()`, which makes it a lot easier to use.

I'm interested in hearing what you guys think. Does it look like a good change?


j08lue:
I actually never used `query_raw`. Why were you using it, @valgur? What could you do with it that could not be done with `query` and `**keywords`?

I also think we could drop the default date values on the unified `query` (i.e. `None` by default). They make most sense in the CLI. We can expect an API user to read the docs. 

We could perhaps raise an error if neither `raw` nor the dates were specified to prevent the user to accidentally query the entire archive.

The benefit of the new `query` having the same parameters as `count` is considerable.

Fernerkundung:
> So, to get the equivalent functionality of query_raw() you can do query(None, None, None, raw="query") 

I'm not a big fan of that design choice. We don't need to cram all functionality into one function. I much prefer our current split and might even argue for splitting it further as to increase the reusage of existing functions. I would argue for `query` staying as a wrapper function for the most common usage and `query_raw` for specialty usage. I don't think your example is unreasonable - it could actually be something helpful to others if it is included in the docs as an example.

> I also think we could drop the default date values on the unified query (i.e. None by default). They make most sense in the CLI. We can expect an API user to read the docs.

Sensible defaults are one of the most important things where I derive data types and valid data ranges from. Personally I only refer to the documentation if the usage of a function is not clear from the docstring and the defaults. My feeling is that when something is not clear from docstring and defaults it is either a very complex function or not well documented.

valgur:
How about keeping both `query_raw()` and `query(..., raw=...)`? This would enable advanced users to run raw queries in a more convenient and robust manner while also making it easy to include raw query strings in normal queries. This would mean that you can achieve pretty much everything with `query()` alone and that the `count()` method's signature can be changed to match `query()`'s.

@j08lue It can be used to construct queries in a bit more convenient manner, though, I don't use it that way: https://scihub.copernicus.eu/userguide/3FullTextSearch#Simple_search_query
I personally like that construct since it allows me to store query parameters as a string in a config file or a DB, so I don't have to deal with serializing and deserializing dictionaries, for example. It's also useful for testing weird non-standard queries, as I have done in some of the previous PRs.

j08lue:
Those are good points, @Fernerkundung.

I agree with you regarding the informative value of defaults. This case is maybe slightly different, though, because the date values can have various format (special relative date string, date string, `datetime`). But defaults still give a very good hint. So scrap that suggestion.

Regarding the docs, I also completely agree that these should only be needed for very complex functions. I actually meant the excellent docstring that @valgur wrote. :-)

I still think `query` having the same call signature as `count` is a benefit. But we can perhaps achieve that without merging `query_raw` into it?

Fernerkundung:
Good suggestion @valgur. To me this sounds like a good compromise. This way we can get the same signatures for `count` and `query`. and using `raw=` is the intermediate step between `query` and `query_raw`.

> I personally like that construct since it allows me to store query parameters as a string in a config file or a DB,

Wouldn't this also be possible with `query_raw`? You'd have to use `format_query` for all strings obviously before storing them in the DB.

valgur:
In the commits above I went ahead with dropping the default values for `initial_date` and `end_date` despite what was discussed earlier.
What convinced me otherwise was trying to do `api.query(aoi, ingestiondate='[... TO NOW]', ... other params ...)` instead of using the main date parameters and failing because of not paying attention to the default date values.
I'd argue that we should not add any parameters to the query by default. This is the most logical behavior in a sense. My main concern with removing the date parameter defaults was removing the effective query size limit from `query()`. I think this has been mitigated by the recent addition of a progressbar for long-running queries. As for the documentation value of the defaults, I think the docstring is pretty clear regarding the values that can be used there.
Do you agree?

I also added support for range values in `query()`, e.g. `query(cloudcoverpercentage=(0, 30))`, and extended the date validation and parsing to all known date types (such as `ingestiondate`) in addition to the main ones. Code reviews are welcome.

Fernerkundung:
I fully support your argument, that we should not inject parameters into the search query that might catch a user off guard. A query without a date specification should be expected to returnthe full size of the archive.

If we are strict about this we should also remove the default of `end_date`. However, I would say this behaviour is expected that a user wants all scenes up to now if they only provide an `initial_date`. However, what about a user only providing the `end_date`? Expected behaviour would be to query the full archive up to this date.

A small add-on that might fit best into this refactor PR: How do you feel about changing the names `initial_date` and `end_date` to something more expressive? I was thinking something along the lines of `sensing_start` and `sensing_end`.

j08lue:
@Fernerkundung you had general restriction on the merge of `query` and `query_raw`. Did you change your mind?

And I thought the result of [our earlier discussion](https://github.com/sentinelsat/sentinelsat/pull/137#issuecomment-309428883) was to keep both?

The question of defaults was secondary.

That all said, I am in favour of your latest proposition, @valgur: removing `query_raw` and also the default arguments on `query` to make it less error-prone to use `raw`.

Regarding your comments, @Fernerkundung:

> If we are strict about this we should also remove the default of end_date.

I think, too, we should have `end_date=None` as a default but set it to `NOW` if `initial_date` is specified but `end_date` is not.

> How do you feel about changing the names initial_date and end_date to something more expressive? I was thinking something along the lines of sensing_start and sensing_end.

I like the clarification that `sensing_start` and `sensing_end` brings. All those dates can be confusing...

Fernerkundung:
> @Fernerkundung you had general restriction on the merge of query and query_raw. Did you change your mind?
> And I thought the result of our earlier discussion was to keep both?

After reading through my earlier reservations I found that they have become mostly obsolete. With the removal of the date defaults the functionality of `query_raw()` just turned into `query(raw=<query>)`. This looks fine to me.

I still maintain, that splitting up `query` into smaller, resuable, manageable functions is a good idea. However, we can still discuss that when the need arises or some time in the future.

valgur:
> A small add-on that might fit best into this refactor PR: How do you feel about changing the names initial_date and end_date to something more expressive? I was thinking something along the lines of sensing_start and sensing_end.

I like the idea. The current names are somewhat vague. The names you suggested are still not entirely clear, though. `sensing_start` and `sensing_end` sound a lot like they might be synonyms to `beginposition` and `endposition` attributes exposed by the API which are related to the orbital slice covered by the image. They appear to be set to the same value for S2 images but are meaningful for S1 images. 
One option for the date parameter might be to merge them and use the newly added interval format. Perhaps something like `sensing_date=("NOW-3DAYS", "NOW")`. I think I would actually still prefer a simpler `date=(...)` and a clarifying sentence in the docstring (that it gets translated to `beginposition`) in that case, though.

>However, what about a user only providing the end_date? Expected behaviour would be to query the full archive up to this date.

> I think we should have end_date=None as a default but set it to NOW if initial_date is specified but end_date is not. That makes it clear that this will not interfere if the user specifies raw.

Good suggestions. We could set them both to `None` by default and change them to values `date(1900, 1, 1)` and `"NOW"` as necessary.

> I still maintain, that splitting up query into smaller, resuable, manageable functions is a good idea. However, we can still discuss that when the need arises or some time in the future.

Sounds reasonable, but what exactly do you have in mind?

Fernerkundung:
> Perhaps something like `date=("NOW-3DAYS", "NOW")`.

Sounds good to me, but how would you expose this in a nice way to the cli? `--date ("20150324", "NOW")` is a bit unwieldy and providing touples or lists is not very common at the command line. Maybe best we just keep `--start` and `--end`?

> what exactly do you have in mind?
Nothing specific, that's why I said we can discuss this should it be necessary in the future. ;) It was more a general mark on the growing functionality and size of `query`, which might hinder maintaineability or new contributions in the future. Let's just keep it the way it is - unless you have something in mind?

valgur:
Regarding the CLI date parameters: I intended to keep them as is, just like the conclusion you came to.

Fernerkundung:
@valgur Would you like to include the refactoring of the date parameters into this PR or rather put it into an issue and include in the next release? Apart from that this PR is looking good to go :+1:

valgur:
Done!

PR #142, opened 2017-07-07
Progressbars

valgur:
Refactored querying code to add a progressbar to long (more than one subquery) queries.

Added a `api._tqdm()` private method that wraps `tqdm()` and adds any values in `api.progressbar_opts` to its keyword args. `progressbar_opts` can be used to slightly customize the progressbar behavior, e.g. disable it entirely or to adjust its update interval. 

The `api._tqdm()` may be overriden by the user to modify the progressbar's behavior, e.g. to replace it with a GUI one. @j08lue suggested this addition [some time ago](https://github.com/sentinelsat/sentinelsat/issues/91#issuecomment-275703332).

I personally saw a need for something like `progressbar_opts` when I saw that the progressbar output started to needlessly fill up my log files when I use `sentinelsat` within [airflow](https://github.com/apache/incubator-airflow) workflows.


sentinelsat/sentinel.py
```diff
@@ -57,6 +57,8 @@ def __init__(self, user, password, api_url='https://scihub.copernicus.eu/apihub/
         self.page_size = 100
         self.user_agent = 'sentinelsat/' + sentinelsat_version
         self.session.headers['User-Agent'] = self.user_agent
+        # Can be modified by the user, e.g. to set to {'disable': True}
+        self.progressbar_opts = {}
```
Fernerkundung: Could you please add this to the docstring? Otherwise it will not appear in the documentation, that these are `tqdm` options and that the progressbar can be disabled.

valgur:
I decided to simplify `progressbar_opts` attribute by replacing it with a `show_progressbars` instance parameter. Enabling/disabling the progressbars is the main choice that we want to expose anyway and the progressbars can still be modified by overriding the `_tqdm()` method, if necessary.

Fernerkundung:
This is looking great!

I'm still uncertain whether the default for the API should be showing progressbars? Either way - everyone can now enable/disable them to their liking.

Issue #115, opened 2017-05-31
Replace positional arguments with keyword arguments, merge CLI functions


valgur:
I propose we replace the CLI positional arguments `<user> <password> <geojson>` with equivalent "optional" arguments (such as `-u,--user`, `-p,--password` and `-g,--geometry`). The `user` and `password` args would still be required arguments in one form or another, of course.

Here are some of the possible benefits of that change:
* It would allow us to implement alternative ways of passing credentials to our CLI, e.g. through [environment variables](http://click.pocoo.org/5/options/#dynamic-defaults-for-prompts), INI or .netrc files (#90).
* Would enable alternative arguments for passing AOI geometries, if need be (#114).
* The user does not need to remember the order of the positional args and can pass arguments in any order. A minor convenience.
* One of ours goals for release v1.0 should be stability and backwards-compatibility of the CLI and API (#108). So now would be a good time for this backwards-incompatible change. This would provide greater flexibility for us in the future, though.

Do you agree?

Fernerkundung:
Completely agree on the proposed changes. I would also add:

- merge `sentinel search` and `sentinel download` into `sentinelsat`

The download of a single scene is the only function of `download`, which can be included into a unified CLI quite easily as a positionaly argument such as `-i, --id`.

j08lue:
I strongly support both of these suggestions.

valgur:
Good idea, @Fernerkundung. I suggest using just `--id` as the keyword, since that's short and consistent with the server-side name. I would also add `--title` to search by the product titles ("S1A_IW_GRDH_...") as well.

Fernerkundung:
I would suggest using `--id` as shorthand for the `UUID` and `--name` to refer `product name` / `file name`, to be consistent with the [OpenHub doumentation](https://scihub.copernicus.eu/twiki/do/view/SciHubUserGuide/5APIsAndBatchScripting#Filtering_the_Products_using_the)

valgur:
I have an alternative idea for the CLI setup: we can keep the `search` and `download` commands but make them share the search parameter arguments. I googled a bit and this can be done pretty easily: https://github.com/pallets/click/issues/108
It would keep the commands and also each commands arguments separate, so arguments that are only relevant to downloading don't pollute the list of arguments for searching, for example. Seems a lot tidier to me.

j08lue:
So you will use something along the lines of [this answer](https://github.com/pallets/click/issues/108#issuecomment-255547347)?

[...]

PR #143, opened 2017-07-09
CLI refactor

Fernerkundung:
Refactoring of the CLI as discussed in #115 

- merged `sentinel search` and `sentinel download` into `sentinelsat`
- replaced positional arguments with `-u, --user`, `-p, --password` (required) and `-g, --geometry`
- `--geometry` still takes the path to a GeoJSON file
- `--uuid` keyword, takes a list of IDs - replacement for `sentinel download`
-   `--name` to search/download by product name - supports wildcards


Fernerkundung:
This PR is ready to merge.

@valgur would you mind giving this a quick review, since you opened #115?

sentinelsat/scripts/cli.py
```diff
@@ -115,21 +114,26 @@ def search(
...
-    products = api.query(wkt, start, end, order_by=order_by, limit=limit, **search_kwargs)
+    if geometry is not None:
+        wkt = geojson_to_wkt(read_geojson(geometry))
+
+    if uuid is not None:
+        uuid_list = [x.strip() for x in uuid.split(',')]
+        products = {}
+        for productid in uuid_list:
+            try:
+                products[productid] = api.get_product_odata(productid)
```
valgur: `get_product_odata` will produce different set of attributes than `query`, which is a slight inconsistency when used with `--footprints`. This is a pretty minor use case, though, which we can safely ignore, though, I think.

sentinelsat/scripts/cli.py
```diff
@@ -115,21 +114,26 @@ def search(
...
+            except SentinelAPIError as e:
+                if 'Invalid key' in e.msg:
+                    logger.error('No product with ID \'%s\' exists on server', productid)
+    elif name is not None:
+        search_kwargs["identifier"] = name
```
valgur: Consider using `_query_names()` here. This would also enable a list of names to be used as input, similar to `--uuid`.

valgur: `_query_names()` does somewhat more than is necessary, though. I would simplify it by moving the 
```python
        # Group the products
        output = OrderedDict((name, dict()) for name in names)
        for id, metadata in products.items():
            name = metadata['identifier']
            output[name][id] = metadata
```
part to `check_files()`. I can take care of it myself, if you wish.

Issue #147, opened 2017-07-21
Under which license is Sentinelsat?


PhML:
In setup.py and README.rst it is GPLv3+.

But the LICENSE file is AGPL.

Which information is the good one?

Fernerkundung:
That's a good question. Originally `sentinelsat` was setup by @willemarcel with AGPL. In the current state references to GPLv3+ are errors.

This might be a good starting point to discuss if we should stick to AGPL or move to another license @willemarcel @j08lue @valgur?

willemarcel:
@Fernerkundung I should have used GPLv3, but I think I copied the license file from another project and didn't notice it was AGPL. Is a message declaring that all the contributors agree with the change enough to move to GPLv3?

Fernerkundung:
>  Is a message declaring that all the contributors agree with the change enough to move to GPLv3?

I don't see why not. We had mixed license information up to now anyway and the difference between AGPL and GPL is only triggered when it is run over a network.

So at the moment we are +2 for GPLv3. Are there other license suggestions or valid reasons not to switch from AGPL to GPLv3+?

j08lue:
Another +1 for GPLv3.

Fernerkundung:
Resolved with #148

PR #148, opened 2017-07-31
change license from AGPL to GPLv3

Fernerkundung:
As discussed in #147


LICENSE
```diff
@@ -629,33 +631,44 @@ to attach them to the start of each source file to most effectively
...
 Also add information on how to contact you by electronic and paper mail.
 
-  If your software can interact with users remotely through a computer
-network, you should also make sure that it provides a way for users to
-get its source.  For example, if your program is a web application, its
-interface could display a "Source" link that leads users to an archive
-of the code.  There are many ways you could offer source, and different
-solutions will be better for different programs; see section 13 for the
-specific requirements.
+  If the program does terminal interaction, make it output a short
```
j08lue: How about this part? Should we actually add that to the CLI?

Fernerkundung: It goes on to say
> when it starts in an interactive mode

Since we do not have an interactive mode I'd say we do not include it. But I am also not opposed to including it - it's just not something I have seen in smaller OSS projects.

j08lue: OK

PR #162, opened 2017-10-11
Add search-by-tile example

### LINKED ISSUE #140: Search Sentinel 2 by tile
Currently, you have to provide an area of interest for `sentinelsat`. But I often know exactly which tile(s) I am interested in. It would be awesome to have a function to download specific tiles, like [`sentinelhub`](https://github.com/sinergise/sentinelhub) has.

As far as I can see, DHuS unfortunately does not have a high-level function to search by tile (see their [Full Text Search](https://scihub.copernicus.eu/userguide/3FullTextSearch) docs).

But you can search for the tile ID in the file name, like `filename:*_TDDABC_*` which however will only work for the [post-2016 single-tile product format](https://earth.esa.int/web/sentinel/user-guides/sentinel-2-msi/naming-convention).

NB: For some reason, combining several `filename` patterns with `OR` like `filename:*_T32UPF_* OR filename:*_T32UMF_*` only returns results for the last tile and `filename:(*_T32UPF_* OR *_T32UMF_*)` does not work at all.

What do you think, should we add a high-level function to the CLI / API to search by tile? It could create the `filename` search pattern and perhaps run several queries when several tiles are provided - unless there is a better way with DHuS that I did not see...

j08lue:
We talked about adding some examples of the API usage to the docs. And that one of the examples could be how to search for Sentinel 2 imagery by tile.

How about this?

Closes #140

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -268,7 +268,7 @@ To search for recent Sentinel 2 imagery by MGRS tile, use the `tileid` parameter
   from collections import OrderedDict
   from sentinelsat import SentinelAPI
 
-  api = SentinelAPI('myuser', 'mypass')
+  api = SentinelAPI('user', 'password')
 
   tiles = ['33VUC', '33UUB']
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -255,6 +255,44 @@ or add a custom handler for ``sentinelsat`` (as implemented in ``cli.py``)
   h.setFormatter(fmt)
   logger.addHandler(h)
 
+More Examples
+-------------
+
+Search Sentinel 2 by tile
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To search for recent Sentinel 2 imagery by MGRS tile, use the `tileid` parameter:
+
+.. code-block:: python
+
+  from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -255,6 +255,44 @@ or add a custom handler for ``sentinelsat`` (as implemented in ``cli.py``)
...
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To search for recent Sentinel 2 imagery by MGRS tile, use the `tileid` parameter:
+
+.. code-block:: python
+
+  from collections import OrderedDict
+  from sentinelsat import SentinelAPI
+
+  api = SentinelAPI('myuser', 'mypass')
```
Fernerkundung: The rest of the docs use ('user', 'password') - would be nice if you could change that to be consistent.

Fernerkundung:
Great addition, especially including the info about the missing metadata regarding the `tileindex` parameter. I did not know that.

Just a heads up - I'm working on a [branch](https://github.com/Fernerkundung/sentinelsat/tree/examples/docs/examples) on collecting examples into a subfolder with Jupyter notebooks that should be included into the docs via sphinx. I'd migrate the existing examples into this after finishing the sphinx integration.

PR #166, opened 2017-10-21
Make checksumming the default behavior

### LINKED ISSUE #129: Make checksumming of downloaded products the default behavior
It is a good practice to verify the integrity of downloaded products and it makes sense to do it by default.
So, let's set `checksum` to `True` in `download()` and `download_all()` and change `--md5` to `--no-md5` in the CLI.
This is a backwards-incompatible change, so we probably want to do implement it in a release that bundles other such changes.

gkeller2:
Fixes #129 

As you can see, I had to modify the tests to remove the uses of the flag `--md5`, which is removed in this code change.

Also updated the README to remove the flag from the description of the CLI tool.

### DIFF
--- a/sentinelsat/scripts/cli.py
+++ b/sentinelsat/scripts/cli.py
@@ -81,11 +81,8 @@ def _set_logger_handler(level='INFO'):
     help="""Create a geojson file search_footprints.geojson with footprints
     and metadata of the returned products.
     """)
-@click.option(
-    '--md5', is_flag=True,
-    help='Verify the MD5 checksum and write corrupt product ids and filenames to corrupt_scenes.txt.')
 @click.version_option(version=sentinelsat_version, prog_name="sentinelsat")
-def cli(user, password, geometry, start, end, uuid, name, download, md5, sentinel, producttype,
+def cli(user, password, geometry, start, end, uuid, name, download, sentinel, producttype,
         instrument, cloud, footprints, path, query, url, order_by, limit):
     """Search for Sentinel products and, optionally, download all the results
     and/or create a geojson file with the search result footprints.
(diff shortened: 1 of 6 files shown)


Fernerkundung:
Thank you for taking time to contribute to this issue @gkeller2 :+1:

Before we merge this I'd like to ask you for some minor changes:

- please also change the default of the [`download`](https://github.com/sentinelsat/sentinelsat/blob/35666648385c07c8e429b1c06add826d3bd8cfe9/sentinelsat/sentinel.py#L387) function as well as the docstring
- remove the MD5 option in the [docs](https://github.com/sentinelsat/sentinelsat/blob/master/docs/cli.rst), just like you did in the Readme
- write the change into the [Changelog](https://github.com/sentinelsat/sentinelsat/blob/master/CHANGELOG.rst)
- and finally: add yourself to the list of contributors at the end of the Readme 😉

gkeller2:
I'm not sure if I updated the Changelog correctly. Please, take a look.

sentinelsat/scripts/cli.py
```diff
@@ -144,12 +141,11 @@ def cli(user, password, geometry, start, end, uuid, name, download, md5, sentine
             outfile.write(gj.dumps(footprints_geojson))
 
     if download is True:
-        product_infos, failed_downloads = api.download_all(products, path, checksum=md5)
-        if md5 is True:
-            if len(failed_downloads) > 0:
-                with open(os.path.join(path, "corrupt_scenes.txt"), "w") as outfile:
-                    for failed_id in failed_downloads:
-                        outfile.write("%s : %s\n" % (failed_id, products[failed_id]['title']))
+        product_infos, failed_downloads = api.download_all(products, path)
```
j08lue: Maybe keep the `checksum=True` here, to make the default behavior visible.

sentinelsat/sentinel.py
```diff
@@ -401,7 +401,7 @@ def download(self, id, directory_path='.', checksum=False):
         checksum : bool, optional
             If True, verify the downloaded file's integrity by checking its MD5 checksum.
             Throws InvalidChecksumError if the checksum does not match.
-            Defaults to False.
+            Defaults to True.
```
j08lue: Why do we write this at all? The default is in the function declaration in this case (would be different if the default was `None`).

gkeller2: You are correct. I agree that there's little benefit to having a line in the method's doc specifying what the default value is, when it's defined in the method's signature. However, I did not write the original documentation -- I merely updated it.

sentinelsat/scripts/cli.py
gkeller2: I find that unnecessary -- specifying a keyword parameter giving it its default value.

j08lue: Yes, you can see it that way, too.

sentinelsat/sentinel.py
Fernerkundung: Documenting the default is optional according to the [numpy docstring standard](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#id5) that we use.

> Optional keyword parameters have default values, which are displayed as part of the function signature. They can also be detailed in the description:

We can remove it. In the case of this PR I'd say updating is the preferred choice. If we want to remove the documented defaults we should do it consistently throughout the codebase.

Fernerkundung:
Thank you for your contribution @gkeller2! I will merge this PR soon and create a release.

PR #168, opened 2017-10-24
Fix unit test regressions

valgur:
It's been a while since the VCR cassettes have been updated. So I did. This broke a bunch of them, tough. About a half of these were due to simple content changes on the server (images appearing or disappearing). The rest were caused by some minor apparent changes in the server-side query handling and required some actual code change. This includes:
* Invalid query strings (such as `xxx:yyy`) no longer raise an error but they return a response with the product count set to `null`, which broke our code. (https://github.com/SentinelDataHub/DataHubSystem/issues/29)
* Unescaped quote symbols no longer cause issues server-side and no longer require a workaround (https://github.com/SentinelDataHub/DataHubSystem/issues/23 has been fixed)
* Query length limit calculation required some updating. The server also returns meaningful error messages in this case now, such as "Request Entity Too Large" and "Request-URI Too Long".


valgur:
The first change raises a small question though. With this PR, invalid query parameters no longer raise an error but return an empty response instead. This reflects the server-side behavior.
Due to the server-side bug, which returns `null` as the product count, we still have this information, implicitly.
Should we continue raising an exception in case of invalid queries? I'm veering on the side of yes, since this is seems more informative and useful to me. Using a server-side bug to implement this is a bit unreliable, though.

j08lue:
I am in favor of raising an error for invalid queries, too. The server will evolve (hopefully) and we will keep adapting our client. If the server at some point stops returning `null`, we will notice, at least when we update the VCRs again, and can reconsider this question then.

Fernerkundung:
+1 on continue to raise the exception.

Even though we state in our contribution guidelines, that "sentinelsat should interface the functions of DHuS where possible" I think a server side bug is clearly not a function.

Issue #176, opened 2017-11-30
geopandas is missed as dependency


skasberger:
When executing to_geodataframe(), the geopandas package was not found on my aws instance. I checked and installed it via pip, and then looked in requirements.txt, where the entry was missing. So I guess this is the problem here, right?

```
geopandas==0.3.0
```

valgur:
We have not included GeoPandas (or even Pandas for that matter) as a requirement since it is a fairly heavy one and is not required outside that single convenience function. In particular, it depends on GDAL, which can be a pain to install on Windows.

We do have [a small note](https://sentinelsat.readthedocs.io/en/stable/install.html#optional-dependencies) about this in the docs, but it can be easy to miss, I guess. Nevertheless, thanks for bringing up a possible issue. I appreciate it.

Fernerkundung:
@valgur Maybe we should could a try/except block into these convenience functions to warn the user about the dependency? Or create an entry into `setup.py` with the dependencies similar to `test` - maybe called `all`?

Not including the dependencies for `to_dataframe` and `to_geodataframe` was a conscious decision by us. However, a small note might not be enough if users are missing it.

I'm reopening this for discussion.

j08lue:
In other packages, I have mostly seen optional dependencies handled the same way as here - with notes in the docs (e.g. [`xarray`](http://xarray.pydata.org/en/stable/installing.html#optional-dependencies)), or maybe several sets of `requirements.txt`, like `requirements[_minimal].txt`, `requirements_full.txt`, `requirements_tests.txt`...

It would be great to find some systematic way of installing optional / lazy dependencies, like with `setup.py`. One problem is that you will hardly ever want to `pip install` `gdal` or `numpy` without having taken care of the libs first - neither on Windows nor on Linux. But that kinda rules out anything but some instructions in the docs...

Fernerkundung:
> you will hardly ever want to pip install gdal or numpy without having taken care of the libs first

Good point.

How about something like this to make it more explicit?

```python
try:
    import geopandas
except ImportError:
    self.logger.warning("This function requires the optional dependency GeoPandas.")
```

j08lue:
You will have to raise an exception, but yes, definitely wrap the `ImportError`.

j08lue:
[...]

PR #177, opened 2017-11-30
increased verbosity regarding optional dependencies

Fernerkundung:
This PR adds more verbose exceptions regarding the optional dependencies for `to_dataframe` and `to_geodataframe`.

The idea is to avoid confusion about the dependencies being optional in case users encounter this prior to reading the docs (which is the standard case I guess). See #176


Fernerkundung:
I've had some trouble mocking the missing dependency for the tests. Turns out monkeypatching them works really well:

`monkeypatch.setitem(sys.modules, "pandas", None)`

PR #179, opened 2017-12-15
Update api.rst

Schlump:
to_geopandas was still used in documentation (redthedocs) instead of to_geodataframe


Fernerkundung:
Thanks for spotting this!

PR #184, opened 2018-01-26
Add ability to set user, password and api_url via environment variables

temal-:
Hi,

thanks a lot for this project.

What nearly kept me from trying it out after installation was the fact that the password has to be supplied on the CLI which made me quite nervous. ;)

Also, setting the API URL via CLI feels quite clumsy, so I added a variable for that too.

The Tests are still working.

Malte


j08lue:
Nice, I did not know that `click` had that feature. That is a very lightweight addition. Do we need test coverage for it? We certainly need documentation.

+1 from me

temal-:
I added a bit of documentation in `README.md` and `docs/cli.rst`. The first commit is something like "hey, you can also set username/password via env variables", while the second commit is more like "it's better to use env variables, so please use them!".

My choice would obviously be to keep both commits (in case I didn't screw up anything else with them) but I can understand if it's too much of a change.

I'm not quite sure about what I did to the ascii art tables. It seemed like the easiest way to add this kind of information, but without a header it might be confusing.

Fernerkundung:
Sorry for replying so late:

I really like the idea and keep being surprised by the power of `click`.
+1 from my side for this PR.

This might be a good point to rethink the naming of the environment variables. I propose to change them from `SENTINEL_*` to `DHUS_*` since we interface the DHuS and the cli commands changed from `sentinel search` to `sentinelsat` a long time ago.

I'll prepare a commit adding to this PR.

j08lue:
> change them from SENTINEL_* to DHUS_*

Sounds good.

Fernerkundung:
Thank you for this PR @temal-.

PR #185, opened 2018-02-08
set user, pass, url for the CLI via env vars

Fernerkundung:
Adding on to #184 by @temal-:

* changed environment variables from SENTINEL_* to DHUS_*
* fixed RST tables
* reverted documentation examples (I think the examples should be self-contained/self-explanatory and therefore include the user and password options)
* changelog


Fernerkundung:
I think this is ready to go. Could you give it a quick look over @j08lue?

Issue #178, opened 2017-12-04
query returns 0 results


Artadys:
Hi,

The line command should be different

I have a problem with `sentinelsat -u myusername -p mypassword -g map-2.geojson --sentinel 2 --cloud 30 --url "https://scihub.copernicus.eu/dhus" `

I built a map-2.geojson above Paris as hereafter (scroll down) file

and I get 0 products

which is not good

Here are the results

```bash
$ sentinelsat -u myusername -p mypassword -g map-2.geojson --sentinel 2 --cloud 30 --url "https://scihub.copernicus.eu/dhus"
Found 0 products
---
0 scenes found with a total size of 0.00 GB
```

Thanks for your attention

PS: map-2.geojson is like that

```
{"type":"FeatureCollection","features":[{"type":"Feature","properties":{},"geometry":{"type":"Polygon","coordinates":[[[1.3842773437499998,49.05227025601607],[2.43896484375,47.87214396888731],[3.71337890625,48.705462895790546],[2.4609375,49.33944093715546],[1.3842773437499998,49.05227025601607]]]}}]}
```

Fernerkundung:
If you leave out a start date using the `-s` option it will query the last 24 hours only. If you add `-s` to your query it works fine, eg. using 2017-11-01 as start and the current date as end.

`sentinelsat -u myusername -p mypassword -g map-2.geojson --sentinel 2 --cloud 30 --url https://scihub.copernicus.eu/dhus -s 20171101`

```bash
Found 29 products
Product cf121829-6703-491f-a25a-9d963ac88554 - Date: 2017-12-02T10:54:21.026Z, Instrument: MSI, Mode: , Satellite: Sentinel-2, Size: 794.43 MB
Product e6161d21-f0c1-4bf5-8e65-1db89e4e2494 - Date: 2017-11-30T11:04:09.027Z, Instrument: MSI, Mode: , Satellite: Sentinel-2, Size: 602.99 MB
...
Product ab4070b2-9b39-45e3-847e-2c94d0261ade - Date: 2017-11-04T10:42:09.027Z, Instrument: MSI, Mode: , Satellite: Sentinel-2, Size: 685.38 MB
---
29 scenes found with a total size of 20.77 GB
```

Artadys:
Great, Thanks I agree….great stuff I learn..

Now I am going to download them.

Thanks




Cristian BEZA <https://www.linkedin.com/in/CristianBeza> | 

[...]

Issue #73, opened 2016-12-03
"Full authentication is required to access this resource"


ibaris:
Hello,

if I try to code a query (python API), I got the error message: 

> SentinelAPIError: # HTTP Status 401 - Full authentication is required to access this resource
> * * *
> **type** Status report
> **message** _Full authentication is required to access this resource_
> **description** _This request requires HTTP authentication._
> * * *
> ### Apache Tomcat/7.0.47 

However, I´m sure that my username and password are correct. 

If I try to search a sentinel scene with the command line, I got the error message: 

> Traceback (most recent call last):[...]
> sentinelsat.sentinel.SentinelAPIError: (HTTP status: 500, code: None)
> # HTTP Status 500 - Server returned HTTP response code: 400 for URL: [...]
> * * *
> **type** Exception report
> **message** _Server returned HTTP response code: 400 for URL: [...]
> **description** _The server encountered an internal error that prevented it from fulfilling this request._
> **exception**
> java.io.IOException: Server returned HTTP response code: 400 for URL: [...]
> **note** _The full stack trace of the root cause is available in the Apache Tomcat/7.0.47 logs._

Here is the script for the Python API:
```
from sentinelsat.sentinel import SentinelAPI, get_coordinates
from datetime import date
import os

api = SentinelAPI('xxx', 'yyy', 'https://scihub.copernicus.eu/dhus/')

WORKDIR    = "C:\...\test"
geofile =  WORKDIR + os.sep + os.listdir(WORKDIR)[0]

# search by polygon, time, and SciHub query keywords

api.query(get_coordinates(geofile), \
          "20151219", date(2015, 12, 29), \
          keywords={"platformname": "Sentinel-1"})
```

Here is the script for the command line:
```
sentinel search -f -s 20140101 -e 20151004 -q "producttype=GRD" -u "https://scihub.copernicus.eu/dhus/" xxx yyy C:\\...\\map.geojson
```

What am I doing wrong?

valgur:
Are your user name and password correct? The examples you provided worked fine for me as long as I use my normal credentials, but when I intentionally uses incorrect ones I got the exact same errors as you did. Does searching the Data Hub work in a browser for you?

With that said, I think we should improve the error message for incorrect credentials (at least for the command line script), since the default server response is a bit vague.

ibaris:
Hello Valgut,

[...]

PR #186, opened 2018-02-08
Add FAQ

martinber:
Hi, I think that a FAQ could be useful, because I was stuck with authentication issues and the answer was on a closed issue (it takes a few days until you can use a new account on apihub).

I don't know if you agree, I looked at some closed issues looking for more problems that people had:

- #178: Query returns 0 results because of time interval (I made that mistake too).
- #178: User asks for something to view the images.
- #73: Authentication issues because of slow account creation on apihub.
- #67 and  #80: 500 time out because of server issues.

So I made a FAQ on README.md with those problems. Keep in mind that maybe I should copy it to the readthedocs documentation, also I'm not a native english speaker so I could have made some mistakes.

Thank you for this useful tool.

### DIFF
--- a/docs/common_problems.rst
+++ b/docs/common_problems.rst
@@ -1,0 +1,41 @@
+.. _common_problems:
+
+Common problems
+===============
+
+.. Using "rubric" directives as titles so they don't show on the TOC
+
+
+.. rubric:: I keep getting *HTTP 401 Unauthorized* messages
+
+This means that the given username/password combination is incorrect. Note that
+if you created your account recently it could take a while (a few days?) until
+you can use that account on *apihub* and therefore *sentinelsat* too. You can go
+`here`__ to test access on the *apihub* endpoint.
+
+__ https://scihub.copernicus.eu/apihub/search?
+
+
+.. rubric:: The query fails with *HTTP 500 connect time out*
+
+Scihub servers are known to have outages due to high demand, try again later.
+
+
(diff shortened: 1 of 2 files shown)


j08lue:
Personally, I hardly ever consult FAQs because they often are formatted in a way that makes it hard to quickly find the problem I have. We could also raise and answer these issues on Stackoverflow and point to that in the docs. That would also make them rank high on Google.

Fernerkundung:
I think in this context it would be more suitable to call them "Known Errors" or "Common Problems" instead of "Frequently Asked Questions".

However, I don't think the Readme is the correct place for this. I'd say these are better listed either:

In the docs:
- pro: works offline and online, can be build by the user
- con: hard to link to Github issues

In the wiki:
- pro: easy to link to issues and PRs
- con: not available offline

If you move this list to either of those locations I'll add my +1, because good documentation matters.

martinber:
There I moved everything to the docs and changed the title to "Common problems". I don't think that the wiki is a good idea because personally I always forget to look there unless there is a link to the wiki on the docs.

j08lue:
> "Known Errors" or "Common Problems"

I kinda like "Known Errors" better, because I don't know how common these problems are - and we don't want to give the (wrong) impression that the library is hard to use. 😄

Fernerkundung:
I've changed the name and merged the documentation of known errors into the master branch.

Thank you for this PR @martinber!

Issue #192, opened 2018-03-29
excluding products already available offline from SentinelAPI.download_all


NiklasKeck:
Good day,
I have been using sentinelsat for quite a while and had to script my way around one limitation:
When using the` SentinelAPI.download_all` with the result of the `query` method (the uuids) as input there is no convenient way to skip specific products without manually filtering this `products` input list. I would love to see this feature in `sentinelsat ` and have implemented it in my local repo. Mabye its a helpful addition for other users as well? 
So here is what I propose:
A new argument to `SentinelAPI.download_all` called `skip_products` that takes as input a list of uuids and/or root folders of products to skip and filters them out from the input argument `products`.

Just to elaborate on why I deem this feature important:
I am working on windows using SARscape to regularly process relatively large stacks of Sentinel-1 data with constant region, sensor, accquisition mode etc (excluding a temporal constraint). Within this workflow I need to unzip and rename the obtained products. I regularly want to download recently published products using the same query. The problem is: Even if I choose the same output directory, all the products that where already downloaded would be downloaded again, since only products with exactly the same filename (local vs. online) are skipped.
Thanks a lot by the way for the great library!

Fernerkundung:
This sounds like a compelling addition to the examples in the documentation. I can see the periodic querying for newly added products as a common use case.

However, I don't think adding `skip_products` adds to the useability of `sentinelsat` for two reasons:

- constructing the list of products to skip is something every user has to do on their own anyway - some may keep a database, others a textfile, etc. Currently we can include the functionality to skip existing, full sized downloaded products in a single directory. Including every other way to check for existing products would be a rather complicated endeavour.

- filtering the products list is a single dict comprehension. I don't think hiding that behind a sentinelsat function adds to the goal of being a Pythonic interface to DHuS.

For instance if you have the downloaded products as unzipped folders one could do something like this:

```python

products = api.query()
skip_products = glob.glob("/Sentinel-1/unzipped/S1A*")

filtered_products = {k:v for k,v in products.items() if v['filename'] not in skip_products}
```

That beeing said, we are sorely lacking more examples for common use cases such as yours in the docs and you are more than welcome to contribute :+1:

NiklasKeck:
That is a good point: There is an enormous amount of different ways one could organize their data. But I think the most common case might be just in plain folders. On windows the path names also have to be changed as they are too long for the file system. This results in the loss of the correct filename so the example you provided would not work anymore.

[...]

PR #194, opened 2018-04-06
Added example 'Exclude Products already available offline from downlo…

NiklasKeck:
Hi there,
I added an example for my use case of sentinelsat as discussed in issue #192.
I tried to make it completelly self contained and therefore it got a little big I guess?! Apart from that this is the first time I got into contact with Sphinx. So I am happy for any comments and critical code review.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -297,6 +297,105 @@ December 2016 onward (i.e. for single-tile products), you can use a `filename` p
 
   kw['filename'] = '*_T{}_*'.format(tile)  # products after 2016-12-01
 
+Exclude Products already available offline from download_all
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``SentinelAPI.download_all()`` automatically skips products that are already in the target directory (current directory
+or as specified in ``directory_path``) if their folder name has not changed. In order to exclude products stored under a
+different location and/or folder name, one has to manually filter the OrderedDict returned from the ``query()``
+method before feeding it to ``download_all()``:
+
+.. code-block:: python
+
+  from sentinelsat import SentinelAPI
+  import xml.etree.ElementTree as ET
+  import os
+  import zipfile
+  import tempfile
+  import time
+  import re
+
+  # make a demo folder:
+  demo_folder = os.path.join(tempfile.gettempdir(),"sentinelsat_exclude_products_demo_{}".format(round(time.time())))
+
+  # On (old) windows systems the maximum path length is restricted to 260 characters. Exceeding this limitation might
+  # result in errors (cp. https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx#maxpath).
+  # user "eryksun" on https://stackoverflow.com/questions/36219317/pathname-too-long-to-open/36237176
+  # offers the following solution (+explanations):
+  if os.name == 'nt':
+    demo_folder = u"\\\\?\\" + demo_folder
+
+  os.mkdir(demo_folder)
+  print("Watch the following folder to see what happens during the demo:\n{}".format(demo_folder))
+
+  # set up a demo query including Sentinel 1 and 2 products
+  api = SentinelAPI('user','password')
+  demo_query = api.query(filename = "S2A_OPER_PRD_MSIL1C_PDMC_20161013T075059_R111_V20161012T161812_20161012T161807.SAFE")
+  demo_query.update(api.query(filename = "S1A_WV_OCN__2SSV_20150526T081641_20150526T082418_006090_007E3E_104C.SAFE"))
+
+  # download the products in the demo_query dictionary
+  downloaded = api.download_all(demo_query, directory_path=demo_folder)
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -297,6 +297,105 @@ December 2016 onward (i.e. for single-tile products), you can use a `filename` p
...
+  print("Watch the following folder to see what happens during the demo:\n{}".format(demo_folder))
+
+  # set up a demo query including Sentinel 1 and 2 products
+  api = SentinelAPI('user','password')
+  demo_query = api.query(filename = "S2A_OPER_PRD_MSIL1C_PDMC_20161013T075059_R111_V20161012T161812_20161012T161807.SAFE")
+  demo_query.update(api.query(filename = "S1A_WV_OCN__2SSV_20150526T081641_20150526T082418_006090_007E3E_104C.SAFE"))
+
+  # download the products in the demo_query dictionary
+  downloaded = api.download_all(demo_query, directory_path=demo_folder)
+  zip_1 = [i for i in os.listdir(demo_folder) if i.endswith(".zip")]
```
j08lue: use `glob.glob(os.path.join(demo_folder, '*.zip'))` instead of that list comprehension around `os.listdir`. Same a couple of lines down.

docs/api.rst
```diff
@@ -297,6 +297,105 @@ December 2016 onward (i.e. for single-tile products), you can use a `filename` p
...
+  for p in zip_1_fullpath:
+      # unzip
+      with zipfile.ZipFile(p, 'r') as zip_f:
+          zip_f.extractall(path = demo_folder)
+      # remove zipfile
+      os.remove(p)
+
+      # rename based on sensor prefix and acquisition time:
+      path_unzipped = p.replace(".zip",".SAFE")
+      prefix,_,acquisition = (re.findall("(S1A|S2A)(.*)(_\d{8}T\d{6})",path_unzipped))[0]
```
j08lue: Sometimes you do have spaces behind comma, as [PEP8](https://www.python.org/dev/peps/pep-0008/) suggests, sometimes you don't. Try using [`flake8`](http://flake8.pycqa.org/en/latest/) or another linter on your code.

j08lue:
Thanks a lot for writing this up so carefully, @NiklasKeck.

> I tried to make it completelly self contained and therefore it got a little big I guess?

I think the length hints at something else: That this is a very special use case that needs a lot of explanation (Windows, your choice of product names, and how you index the stored products by XML or PDF). With all this logic, I think this is actually not suited for the docs.

In my opinion, the examples in the docs should mostly illustrate briefly how to use the API out of the box, without adding too much special logic. But the other maintainers might disagree.

Great that you want to share your solution for storage and indexing, since it might be useful to others, too. But I really don't know where the best place for this is - I think it is neither the source code (as discussed in #192) nor the docs. Maybe a question on Stackoverflow that you answer yourself? [SO does encourage you to do that](https://stackoverflow.com/help/self-answer).

docs/api.rst
NiklasKeck: Thanks a lot for this comment! I can definitly put this to use in the future.

docs/api.rst
NiklasKeck: Thank you for the comment. I will watch out those inconsistencies and try to make my IDE do the same :)

NiklasKeck:
Hi @j08lue thank you for your code review. It is true that this is just one of many use cases, but I think quite a common one as the following requirements do not seem like a very rare case. User that:
1. use windows,
2. want to download new products with the same query as soon as they are available without redownloading the older products,
3. use software that does not work with the original zip files (unlike e. g. https://github.com/senbox-org/),
4. don't want to bother with writing additional files to disk to keep track of the original naming/uuids.

(The specific renaming scheme I used in the example is off course random and not important)

I don't know if this is better than moving this to Stackoverflow, but maybe there could be a 'use cases' section in the docs, which always starts with requirements tags (e. g. windows, regularly download new products, need to unzip) and then gives an example implementation (such as my PR). This would help those who have this specific use case without bloating the Python API or the Command Line Interface sections but still keep this use case within the official docs. 
Keeping it with the original docs would save the user time in my opinion. At least it would have saved me time, as searching on SO is usually a step that I do only when I don't understand how and why something does not work from the original docs and alreay have enough insights to enter the right question/query into google, which takes time.

Fernerkundung:
>In my opinion, the examples in the docs should mostly illustrate briefly how to use the API out of the box, without adding too much special logic. But the other maintainers might disagree.

I fully agree. The documentation should document how to use the API and include generic examples. If an example needs more than two pre-requesites I don't think it is generic enough to to be documentation of the API.

However, I see a value in (and the work put into) this example as it outlines how people can use the API to build their own logic depending on their use cases.

As for the best place to put these use cases - how about a section in the Github Wiki? This way it is linked to the project but does not bloat the code base and is easier to find than a question on StackOverflow or a blog post. As this project values good documentation I'd also propose to include people contributing to the Wiki sections as contributors in this projects list.

NiklasKeck:
Thank you @Fernerkundung and @j08lue for your comments. I edited the wiki to include my use case. Please feel free to further edit my contribution according to your taste.
I am therefore closing this pull request. Thank you for the great package!

PR #202, opened 2018-06-06
Fix order_by not being included in subqueries

### LINKED ISSUE #200: number of products returned by _parse_opensearch_response(products) is different to incoming products list
_parse_opensearch_response(products) reformats the query result in query() 
set but it seems that there is a paging problem so the length of the orderedDict is in most cases exactly 100 or sometimes below (depending on length of products list). 
As 100 is the configured self.page_size it seems to me that the last page get somehow ignored.

valgur:
Fixes #200.

Also modified `count()` slightly to make its function signature fully compatible with `query()`, so exactly the same arguments can be used for both of them. Irrelevant arguments (`order_by`, `offset` and `limit`) are dropped, so they would not get erroneously included as query keywords.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ Added
 Changed
 ~~~~~~~
 * remove hard coded product type list from cli (#190, @lenniezelk)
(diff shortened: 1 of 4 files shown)


Fernerkundung:
Great work on spotting the issue and providing a fix on such short notice :+1:

PR #207, opened 2018-06-21
Block all network traffic when using VCR cassettes for unit tests

valgur:
Disable the `socket.socket` interface to block all network traffic when running unit tests against prerecorded queries. This should make writing unit tests a bit easier by explicitly enforcing our convention of not doing any network requests by default when running unit tests.

I wanted to add this very minor enhancement already about a year ago, but the `pytest-socket` extension needed [some tweaking](https://github.com/miketheman/pytest-socket/pull/1) first for this to work in a reasonable manner.

### DIFF
--- a/CONTRIBUTE.rst
+++ b/CONTRIBUTE.rst
@@ -93,7 +93,10 @@ To run the tests
   py.test -v
 
 
-By default, prerecorded responses to Copernicus Open Access Hub queries are used to not be affected by its downtime. To allow the tests to run actual queries against the Copernicus Open Access Hub set the environment variables and add ``--vcr disable`` to ``py.test`` arguments.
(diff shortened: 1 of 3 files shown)


Fernerkundung:
Should we document this somewhere, so people writing new tests are not caught off-guard? Or do you think the documentation of the usage of VCR and how to record new responses is enough?

I tend towards the latter.

valgur:
Good point. The exception raised in case of inadvertent network accesses is not really super clear about its intended purpose (`pytest_socket.SocketBlockedError: A test tried to use socket.socket.`), so I added a small explanation of this feature to CONTRIBUTE.rst.

Fernerkundung:
Perfect :+1:

PR #208, opened 2018-06-23
Tidy up docs, remove requirements.txt

valgur:
* Replace Numpydoc with Napoleon for better-looking docstring formatting. Numpydoc produces pretty messy-looking parameter lists right now for some reason.
* Use code order instead of alphabetical for class member list ordering in API reference.
* Clean up docstrings to produce nicer HTML documentation.
* Make sure all cross-reference links in docs work.
* Add -h in addition to --help argument to CLI.
* Rename 'Known errors' -> 'Common issues'. I think 'Known errors' sounds too much like 'yet-unresolved sentinelsat bugs' while the page itself describes common Data Hub problems.
* Improved `test_docs` to automatically find all *.rst files.
* Got rid of all warnings and errors in building docs.
* ~Make building docs fail if any warnings or errors are present.~ Reverted this. I think it would be too much of a hindrance.


valgur:
By the way, @Fernerkundung, could you replace the ReadTheDocs [GitHub service](https://github.com/sentinelsat/sentinelsat/settings/installations) with a [webhook](https://docs.readthedocs.io/en/latest/webhooks.html#github) as the services are being deprecated?

setup.py
```diff
@@ -40,18 +40,19 @@
               'pytest',
               'requests-mock',
               'vcrpy',
-              'rstcheck'
+              'rstcheck',
+              'sphinx'
```
Fernerkundung: Napoleon requires sphinx >= 1.3 - think we should include this version dependencance.

Fernerkundung:
@valgur I removed the RTD service and added a webhook.

valgur:
> Napoleon requires sphinx >= 1.3 - think we should include this version dependencance.

I was not sure whether to place that in `requirements.txt` or `setup.py`, so I took the advice at https://packaging.python.org/discussions/install-requires-vs-requirements/ and placed it inside `setup.py`.

Since the [test] installation target now also includes `sphinx`, there is not much need for having a separate [docs] target, so I merged them into a single [dev] (and [dev34]).

Also, based on the advice given in the above link, it seems that there is no point in having a separate `requirements.txt` -- we do not need it to pin our versions or specify an alternative source of packages. So I suggest we remove it altogether? What do you think?

Fernerkundung:
:+1: on merging the extra installation entry points.

I'm :-1: on removing `requirements.txt`. We have never used it but I see value in a de-facto standard that people expect and look for to get a easy to read overview of the dependencies.

j08lue:
> on removing requirements.txt

I do not have a strong opinion on this, but I think it is fine to drop `requirements.txt`. Removes some clutter and people who want to know what the dependencies are will look for them in `setup.py`, too, I'd say.

Fernerkundung:
With a vote of 2:1 the removal of the `requirements.txt` can be regarded as a passed proposition.

Anything else you want to add @valgur? Otherwise this looks ready to merge. The improved docs look a lot better than with numpydoc - great job!

Issue #169, opened 2017-10-30
Protect spaces in query values


PhML:
I used Sentinelsat with https://coda.eumetsat.int/ (The Sentinel 3 Marine 
Copernicus Data Access) and the extra keyword `timeliness` which expects values with spaces like `Near Real Time`.
Unfortunately when Sentinelsat does the query, such values are not protected, leading to a query error response from the server.
I tried to modify Sentinelsat to surround each value with double quotes `"` in queries but it failed when values are range values like `[NOW-1DAY TO NOW]`. It also failed with `filename` keyword and an asterisk `*` in the value like `S3A_OL_2*`. I eventually added this condition in the `format_query` method:
```python
            if " " in value and not value.startswith("["):
                query_parts.append('{}:"{}"'.format(attr, value))
            else:
                query_parts.append('{}:{}'.format(attr, value))
```
Instead of just:
```python
            query_parts.append('{}:{}'.format(attr, value))
```
On queries I made, it worked but as I find it very strange that I can't use double quotes everywhere, I ask for a discussion on this behavior.

valgur:
You raise a good issue, this opens a whole can of worms. 😄 

I dug into the Apache Lucene query syntax that the server uses ([here](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/query-dsl-query-string-query.html#query-string-syntax)'s a good reference). There are two ways spaces can be supported in keywords:
* Quotes: `timeliness:"Non Time Critical"`. This has the downside that queries like `identifier:"S1*"` and  `relativeorbitnumber:"[100 TO 120]"` will not work.
* Escaping: `timeliness:Non\ Time\ Critical`. This has the downside that we have to take care to not escape spaces that are used as part of the syntax, e.g. `relativeorbitnumber:[100 TO 120]`.
* Using parentheses (`timeliness:(Non Time Critical)`)  also seemingly works, but this is actually equivalent to `timeliness:(Non AND Time AND Critical)` and `timeliness:Critical AND timeliness:Time AND timeliness:Non`.

So whichever approach we take we would have to check the content of the keyword first. I would choose the escaping approach and check for the presence of square brackets. In case of quoting we would also have to check for the presence of wildcards, which is a bit more extra work.

Mostly unrelated, but going over the query syntax spec I discovered some other interesting features I was not yet aware of:
[...]

PR #210, opened 2018-06-24
Add support for single-sided ranges, check for duplicate keywords

valgur:
Date ranges can now also use `*` for single-sided ranges, e.g. `query(date=('*', 'NOW-1YEAR'))`.

Improved `test_format_date` by checking that all of the tested date strings are also accepted/rejected on the server as expected.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,7 @@ All notable changes to ``sentinelsat`` will be listed here.
 
 Added
 ~~~~~
-*
+* Date ranges now also support '\*' for single-sided ranges, e.g. ``query(date=('*', 'NOW-1YEAR'))``.
 
 Changed
 ~~~~~~~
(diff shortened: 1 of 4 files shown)


valgur:
@j08lue This is the server-side syntax for it, which I noticed exists in [#169](https://github.com/sentinelsat/sentinelsat/issues/169#issuecomment-340433778). I wanted to make sure that we at least do not block it in our client, since it's useful. But adding `None` as a synonym for it makes a lot of sense for a Python API and a great idea. Thanks! I'll add it.

valgur:
I updated the API to allow single-sided ranges to be used for non-date ranges as well, also allowing `None` to be used. In case both bounds are set to unlimited, the keyword will be removed from the query.

I also added a check for the presence of duplicate keywords in a query. This can happen due to capitalization (`beginposition` and `beginPosition`) or mixing function parameters and `**keywords` (`date` and `beginPosition`).

PR #212, opened 2018-06-26
fix readthedocs requirements

Fernerkundung:
RTD was recursively pulling in `requirements.txt` - after removal the docs build failed.


valgur:
Hmm, I did not notice that RTD is using `requirements.txt`. In that case I would revert back to keeping it. I suppose you agree?

Fernerkundung:
see #213

PR #213, opened 2018-06-26
Restore requirements.txt

valgur:
I did not notice that a separate requirements.txt file is useful to have for ReadTheDocs builds, so I suggest we restore it. (related to #208 and #212)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,7 +16,7 @@ Added
 
 Changed
 ~~~~~~~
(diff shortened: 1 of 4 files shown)


Fernerkundung:
In that case we can remove `docs/readthedocs-requirements.txt` and I will point RTD straight to `requirements.txt`

I forgot about the recursive inclusion into the rtd-requirements. Reason I included it was the use of numpydoc, which did not appear in the requirements file and RTD had no way to access different setup.py entry points.

PR #214, opened 2018-07-01
add Python 3.7 to test matrix

Fernerkundung:


Fernerkundung:
Looks like work for Python 3.7 support in Travis is still ongoing. See: https://github.com/travis-ci/travis-ci/issues/9069

valgur:
There's now a provided solution in the thread you linked to: https://github.com/travis-ci/travis-ci/issues/9069#issuecomment-401828871

Would you please update the pull request based on that and merge, @Fernerkundung?

Fernerkundung:
Python 3.7 works as a target now. The tests fail due to the `geopandas` dependency `fiona` failing to build. @valgur can you see from the error message if we can fix that or have to file an issue with `fiona`?

I don't mind if you take over from here if you want - I currently have no time to investigate further.

valgur:
Also `pyproj` and `pyyaml`... We should try again in a month or so and hope that the dependencies have upgraded themselves to Python 3.7.

Issue #215, opened 2018-07-05
Reconnect when drop to minimal speed


c040340:
Hi just found that for bad network circumstance, the server is going to stop responding and the downloading procedure will stuck. 

To handle this I just made some modification on your code, I added an "timeout" flag to your downloading session in https://github.com/sentinelsat/sentinelsat/blob/fea079487dce41ccfe049a6e856a8287b77d8fe3/sentinelsat/sentinel.py#L741 and use "while" function to make _download() function running again and again. 

This worked very will for bad network circumstance.

Fernerkundung:
How do you detect a timeout - i.e. what is the cut-off when a download is considered to be timed out vs. running as expected?

c040340:
Since you have already set the chuck size to 2*20 (2MB or 1MB, in line 745), you can just set "timeout" to 2000 (unit is second) and this means when the download speed is lower than 1KB/s we are going to re-download the chunk. 

I just found this is very helpful for my network. 

Cheers,

Fernerkundung:
That does sound like a reasonable default for a lower boundary. Would you mind preparing a pull request for this?

I think it would be useful to set this as an parameter of the `SentinelAPI` class so users with faster/slower connections can set it accordingly. What's your take on this @valgur @j08lue ?

c040340:
Yeah, I am glad to do that. Do you need to invite me to the group before I make the changes?

j08lue:
> I think it would be useful to set this as an parameter of the SentinelAPI class so users with faster/slower connections can set it accordingly.

Sounds like the perfect solution for this. This will be a good addition, @c040340, thanks.

j08lue:
The change we envisioned above has been implemented in #259. I am taking the liberty to close this.

PR #217, opened 2018-07-17
lowest speed added

c040340:
See issue #215

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -755,7 +755,7 @@ class SentinelAPI:
                     progress.update(len(block_data))
             return md5.hexdigest().lower() == checksum.lower()
 
-    def _download(self, url, path, session, file_size):
+    def _download(self, url, path, session, file_size, speed_boundary_in_kb=1):
         headers = {}
         continuing = exists(path)
         if continuing:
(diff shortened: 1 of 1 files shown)


sentinelsat/sentinel.py
```diff
@@ -755,7 +755,7 @@ def _md5_compare(self, file_path, checksum, block_size=2 ** 13):
                     progress.update(len(block_data))
             return md5.hexdigest().lower() == checksum.lower()
 
-    def _download(self, url, path, session, file_size):
+    def _download(self, url, path, session, file_size, speed_boundary_in_kb=1):
```
Fernerkundung: * please change the name from "speed_boundary_in_kb" to "timeout"
* timeout should be a class parameter that is provided with the class, not a function parameter
* documentation is necessary in the class docstring that the timeout refers to seconds per 1MB chunk

sentinelsat/sentinel.py
```diff
@@ -764,20 +764,26 @@ def _download(self, url, path, session, file_size):
...
-            mode = 'ab' if continuing else 'wb'
-            with open(path, mode) as f:
-                for chunk in r.iter_content(chunk_size=chunk_size):
-                    if chunk:  # filter out keep-alive new chunks
-                        f.write(chunk)
-                        progress.update(len(chunk))
-                        downloaded_bytes += len(chunk)
-            # Return the number of bytes downloaded
-            return downloaded_bytes
+        while True:
```
Fernerkundung: * avoid the `while True` construction, as this will retry indefinetly
* we have a retry logic in place in the `download_all` function with `max_attempts`: https://github.com/sentinelsat/sentinelsat/blob/master/sentinelsat/sentinel.py#L500
* `timeout` should be a parameter of `download_all`, just like `max_attempts`, with a default value

This construction would break the retry logic.

Fernerkundung:
A test for the `timeout` would be nice. However, I am not quite sure how we could mock a slow connection. This might require fiddling with controllable sockets (see https://stackoverflow.com/questions/1833563/simple-way-to-simulate-a-slow-network-in-python).

Fernerkundung:
Deferring in favor of #256

Issue #223, opened 2018-08-10
geojson_to_wkt fails on valid geojson object with 'geometry' key at top level


scottstanie:
I'm trying to use the api for a simple Polygon that looks like this:
```
{
  "type": "Feature",
  "properties": {},
  "geometry": {
	"type": "Polygon",
	"coordinates": [
	  [
		[
		  -103.46099853515624,
		  30.95876857077987
		],
		[
		  -102.72491455078125,
		  30.95876857077987
		],
		[
		  -102.72491455078125,
		  31.85423130442635
		],
		[
		  -103.46099853515624,
		  31.85423130442635
		],
		[
		  -103.46099853515624,
		  30.95876857077987
		]
	  ]
	]
  }
}
```

However, the code for the `geojson_to_wkt` function fails on a key error since it does this:

```python
if 'coordinates' in geojson_obj:
    geometry = geojson_obj
else:
    geometry = geojson_obj['features'][feature_number]['geometry']
```

It seems like it should be able to handle this type of geojson, unless I've missed something in the documentation forbidding it? The first example on the homepage for geojson also fails due to this problem: http://geojson.org/

scottstanie:
Let me know if you also would like me to add a test for this new loading geojson type.

j08lue:
> Let me know if you also would like me to add a test for this new loading geojson type.

yes, please

Fernerkundung:
Closed with #224

PR #224, opened 2018-08-10
Add check for 'geometry' key in geojson_to_wkt

scottstanie:
This solves issue #223 , allows for geojson objects with the coordinates nested one level deep that are still simple Polygons.


j08lue:
Good idea to support this, @scottstanie.

Seems like the CI test failures were not because of your code (something with `socket`). 

But can you please add a test for this?

scottstanie:
@j08lue Alright test added by just taking a layer out of `map.geojson`

PR #225, opened 2018-08-10
fix `area_relation` docstring options

scottstanie:
Doesn't match the current options, and `Intersection` throws an error

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,6 +29,7 @@ Fixed
 ~~~~~
 * Spaces in query parameter values are now escaped by adding a backslash before them, where appropriate. (#169, #211)
 * Fixed some CLI errors not returning with a non-zero exit code. (#209)
+* Fixed typo for ``area_relation`` query parameter documentation from ``'Intersection'`` to ``'Intersects'``. (#225)
 
 [0.12.2] – 2018-06-20
 ---------------------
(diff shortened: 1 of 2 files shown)


j08lue:
Can you link to the DHuS docs / changelog or so where this change is mentioned?

scottstanie:
Hmm not sure I know how to find it in the DHuS docs, but it seems like it never changed here from "intersection" to "intersects"- it might have been a typo perhaps?
The only mention in this repo's changelog in the commit that added it:
https://github.com/sentinelsat/sentinelsat/commit/76582dfd87a4edc0ed062cc2f960f6b82601443b

And in that one it's also slightly typo-ed- "Interesects", but the actual option never seemed to be "intersection".

Did you want me to add a line to the changelog, or just try and find where it might have changed from "intersection" to "intersects"?

j08lue:
Ah, I thought this was a recent change in DHuS and you got it from their changelog or so. So it was just a bug on our end then.

> Did you want me to add a line to the changelog

That, too, please. In the `[master]` section under 'Fixed'.

scottstanie:
Alright added to CHANGELOG

j08lue:
Thanks @scottstanie.

@valgur Do you know why we get [these test failures under Python 3.5](https://travis-ci.com/sentinelsat/sentinelsat/jobs/140961453#L985)? 

```
pytest_socket.SocketBlockedError: A test tried to use socket.socket.
```

I first thought we might need to update the cassettes, but why would that be an issue only under Python 3.5...?

valgur:
The unit test failures seem to have been fixed in the minor VCR.py usage adjustments (in the `products` and `raw_products` fixtures) in #230. I expect the build failure encountered here to be resolved once that PR gets merged.

PR #230, opened 2018-09-05
Update check_query_length(), Unicode support

valgur:
The server-side query string handling has changed a bit so we need to update the calculation logic in `check_query_length()`. Related to #228.


valgur:
I also added Unicode support for search queries. The server parses query strings as Latin-1 by default, but can be made to accept UTF-8 by setting a specific header.
I know there is hardly ever a need for Unicode symbols in queries, but it does not hurt to have this support either.

valgur:
The unit test for Unicode support is currently disabled due to a VCR.py bug. The test can be enabled again once https://github.com/kevin1024/vcrpy/pull/386 gets pulled and included in a release.

Issue #227, opened 2018-09-03
Error no file or directory "search_polygon.geojson"


Overature:
I am trying to install the sentinet sat api as per [these](https://sentinelsat.readthedocs.io/en/stable/ in) instructions. 

However the code is exactly the same as the powerful api section from that link with my password and username, But I get:

Error no file or directory "search_polygon.geojson"

as a error, can anyone help? 

Seems to be working fine from the command line. 

Thanks

j08lue:
The example in the docs is **not** self-contained. It expects that you have a file `search_polygon.geojson` in the directory you are running from.

```python
footprint = geojson_to_wkt(read_geojson('search_polygon.geojson'))
```

If you instead point to a real GeoJSON file, e.g. one you make on http://geojson.io, it will work.

```python
footprint = geojson_to_wkt(read_geojson('/path/to/my/map.geojson'))
```

Overature:
thanks for your help, if i download a product it works fine.

j08lue:
Please close this issue then. That is usually the original poster's privilege. 😄

PR #235, opened 2018-10-08
Clarify API usage example in the docs closes #229

### LINKED ISSUE #229: Clarify API usage example in the docs
Based on #227, we should make it more clear in the docs that `search_polygon.geojson` is a file you must provide.

psal93:
Added GeoJSON clarification to docs api.rst, cli.rst and index.rst


docs/api.rst
```diff
@@ -47,6 +47,9 @@ Quickstart
 Valid search query keywords can be found at the `Copernicus Open Access Hub documentation
 <https://scihub.copernicus.eu/userguide/3FullTextSearch>`_.
 
+Search areas (i.e. ``map.geojson`` ) are provided as GeoJSON polygons, which can be created with
```
j08lue: No, maybe we should clarify this differently here: The `SentinelAPI.query` function actually takes the area of interest as WKT, not GeoJSON. 

The workflow `GeoJSON file > read to string > convert to WKT > pass to api.query` is pretty clear in this line `footprint = geojson_to_wkt(read_geojson('map.geojson'))`. 

So the primary thing to clarify here is that `map.geojson` is a path to a file named that, which was the source of issue #227.

docs/cli.rst
```diff
@@ -13,7 +13,7 @@ password to access the Copernicus Open Access Hub.
 
 .. code-block:: bash
 
-  sentinelsat -u <user> -p <password> -g <geojson>
+  sentinelsat -u <user> -p <password> -g <search_polygon.geojson>
```
j08lue: yes

docs/cli.rst
```diff
@@ -23,9 +23,9 @@ For convenience and added security you can set the username, password and DHuS U
   export DHUS_PASSWORD="<password>"
   export DHUS_URL="<api_url>"
 
-  sentinelsat -g <geojson>
+  sentinelsat -g <search_polygon.geojson>
 
-Search areas are provided as GeoJSON polygons, which can be created with
+Search areas (i.e. ``search_polygon.geojson`` ) are provided as GeoJSON polygons, which can be created with
```
j08lue: Strictly speaking, we also support `MultiPolygon` and `Point`, not just `Polygon`s. So instead of `GeoJSON polygons`, maybe write `GeoJSON files` here.

docs/index.rst
```diff
@@ -11,7 +11,7 @@ It offers an easy-to-use command line interface
 
 .. code-block:: bash
 
-  sentinelsat -u <user> -p <password> -g <search_polygon.geojson> --sentinel 2 --cloud 30
+  sentinelsat -u <user> -p <password> -g </path_to_data/search_polygon.geojson> --sentinel 2 --cloud 30
```
j08lue: A matter of taste, but maybe call it `/path/to/search_polygon.geojson` instead?

j08lue:
By the way, please add one of the [magic GitHub phrases](https://help.github.com/articles/closing-issues-using-keywords/) to your PR description, so the original issue (`#229`) is automatically closed when we merge this.

j08lue:
Thanks a lot, @psal93. I think this is ready to merge now.

PR #240, opened 2018-10-11
query for multiple indentifiers using OR

### LINKED ISSUE #156: Allow --name to take multiple inputs
The `--name` parameter could be fairly easily extended to take a list of product names as input. Perhaps as a comma separated list.

j08lue:
Closes #156 

This makes it possible to pass multiple identifiers to `sentinelsat --name` as a comma-separated string, as envisioned in the linked issue.

For this, I turned off whitespace escape for query strings in parentheses `(...)`. Do you see any problems with that, @valgur and @Fernerkundung?

### DIFF
--- a/sentinelsat/scripts/cli.py
+++ b/sentinelsat/scripts/cli.py
@@ -19,6 +19,16 @@ def _set_logger_handler(level='INFO'):
     logger.addHandler(h)
 
 
+class CommaSeparatedString(click.ParamType):
+    name = 'comma-string'
+
+    def convert(self, value, param, ctx):
+        if value:
+            return value.split(',')
+        else:
+            return value
+
+
 @click.command(context_settings=dict(help_option_names=['-h', '--help']))
 @click.option(
     '--user', '-u', type=str, required=True, envvar='DHUS_USER',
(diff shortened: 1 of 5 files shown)


Fernerkundung:
I don't see a problem with removing whitespace escape. However, I would like to divert that question to @valgur as he has the most experience with the query strings.

j08lue:
Just to clarify - we already leave whitespace alone if it is in `[]` or `{}`. Now I just added `()`, too. I could probably also use `[... OR ... OR ...]` and would not need that change, but I did not test that.

sentinelsat/scripts/cli.py
```diff
@@ -19,6 +19,16 @@ def _set_logger_handler(level='INFO'):
     logger.addHandler(h)
 
 
+class CommaSeparatedString(click.ParamType):
```
valgur: This is a really nice touch! 👍

sentinelsat/sentinel.py
```diff
@@ -164,7 +164,7 @@ def format_query(area=None, date=None, raw=None, area_relation='Intersects',
             # Escape spaces, where appropriate
             if isinstance(value, string_types):
                 value = value.strip()
-                if not any(value.startswith(s[0]) and value.endswith(s[1]) for s in ['[]', '{}', '//']):
+                if not any(value.startswith(s[0]) and value.endswith(s[1]) for s in ['[]', '{}', '//', '()']):
```
valgur: I'm not sure why I did not include `()` in this list. This should be absolutely fine.

Issue #131, opened 2017-06-09
Make exceptions more specific


valgur:
I first suggested this change in #108.

I would split `SentinelAPIError` into a couple of more specific ones:

  - `SentinelAPIError`: for relaying API responses that return an exception, e.g. for invalid query parameters. The response is expected to contain the `cause-message` in that case (as in #107).
  - `UnauthorizedError`: for cases where invalid credentials were provided or when the server does not accept them yet for whatever reason (a common issue for new users, I think).
  - `ServerError`: for cases when the server is unreachable, under maintenance or the response is just unparsable garbage.

This would enable different error cases to be caught more elegantly. E.g. you probably just want to retry your query after some delay in case of `ServerError`.

gkeller2:
Started working on this today. I'll have an early version of the change soon to discuss.

Andrey-Raspopov:
Is this active? If it is, I can try to divide them.

Fernerkundung:
@Andrey-Raspopov I think we can reasonably assume that @gkeller2 isn't working on it anymore a year later. Go for it 😃

Andrey-Raspopov:
Are there any default project credentials for scihub? I've registered for API access a week ago and still can't run a query.

Fernerkundung:
There are no default project credentials, as account sharing is forbidden by the SciHub terms and conditions.

The propagation of accounts to the apihub used to take a while. If you test with `https://scihub.copernicus.eu/dhus/` the registered accounts work right away (just trested with a new one).

Andrey-Raspopov:
Yeah, it works there, and it worked since day one, so I suppose I'll just wait more:)

Fernerkundung:
No need to wait - the API for both endpoints is identical. You can test with '/dhus' instead of `/apihub` locally without any issues.

dwlsalmeida:
Is someone currently working on this or is it still an issue?

I'd like to work on that, seems like a good first issue.

valgur:
It's definitely up for grabs, go ahead!

Creating realistic mock responses for unit tests might be somewhat tricky, especially for `ServerError`, but the existing tests should provide some useful examples. You will likely need to use something like [`requests-mock`](https://requests-mock.readthedocs.io/en/latest/index.html) to set up test cases for `ServerError`.

dwlsalmeida:
I have some of the code ready, but so far I've been unable to run the tests.

Here's the stacktrace for `py.test -v -vcr disable`

[...]

PR #242, opened 2018-10-16
Change several variables to comply with PEP-8

Andrey-Raspopov:
Several variables' names were shadowing built-ins or higher-level variables. While not crucial and critical in current situation, it can become troublesome later.


sentinelsat/sentinel.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, user, password, api_url='https://scihub.copernicus.eu/apihub/
         self._last_query = None
         self._last_response = None
 
-    def query(self, area=None, date=None, raw=None, area_relation='Intersects',
```
j08lue: No, please don't change the query parameter name `date`. Rather do `import datetime` and `datetime.date`, which we should be doing anyways, IMHO.

sentinelsat/sentinel.py
```diff
@@ -378,7 +378,7 @@ def to_geodataframe(products):
         df.drop(['footprint', 'gmlfootprint'], axis=1, inplace=True)
         return gpd.GeoDataFrame(df, crs=crs, geometry=geometry)
 
-    def get_product_odata(self, id, full=False):
+    def get_product_odata(self, product_id, full=False):
```
j08lue: yes

sentinelsat/sentinel.py
```diff
@@ -835,19 +835,19 @@ def geojson_to_wkt(geojson_obj, feature_number=0, decimals=4):
     else:
         geometry = geojson_obj['features'][feature_number]['geometry']
 
-    def ensure_2d(geometry):
-        if isinstance(geometry[0], (list, tuple)):
-            return list(map(ensure_2d, geometry))
+    def ensure_2d(_geometry):
```
j08lue: why? I have never seen `_private` parameters to functions...

sentinelsat/sentinel.py
```diff
@@ -6,7 +6,7 @@
 import re
 import shutil
 import warnings
-import xml.etree.ElementTree as ET
+import xml.etree.ElementTree as ElemTree
```
j08lue: Hmm. I'd say you might as well `from xml.etree import ElementTree`.

sentinelsat/sentinel.py
```diff
@@ -675,8 +675,8 @@ def check_files(self, paths=None, ids=None, directory=None, delete=False):
         paths = paths or []
         ids = ids or []
 
-        def name_from_path(path):
-            return splitext(basename(path))[0]
+        def name_from_path(base_path):
```
j08lue: No, please, `path` is absolutely legal to use here. No one in their right might should use `from os import path`. 😄

sentinelsat/sentinel.py
j08lue: Actually, even in the [Python docs](https://docs.python.org/3/library/xml.etree.elementtree.html) they use `ET`.

sentinelsat/sentinel.py
Andrey-Raspopov: Agreed, I'll rewrite it that way then. But there are several usages of datetime.datetime.<something>. It seems that these imports won't be really nice anyway.

j08lue: I prefer `datetime.datetime`, because just `datetime` is so ambiguous. Weird that they did not call it `datetime.Datetime`, since it is a class, isn't it?

sentinelsat/sentinel.py
Andrey-Raspopov: Yeah, I was struggling with the name and just used underscore, which misses the point.

j08lue: I see. What does `geometry` conflict with?

sentinelsat/sentinel.py
Andrey-Raspopov: I renamed it not because of os.path (which is imported in another way), but because we use "path" as a variable name in the outer function.

sentinelsat/sentinel.py
Andrey-Raspopov: Agreed

Andrey-Raspopov:
I've reverted some changes and ensured that tests are not broken by this commit

sentinelsat/sentinel.py
Andrey-Raspopov: Absolutely, I suppose somewhere deep in the python community we can find heated arguments about this naming:)

sentinelsat/sentinel.py
Andrey-Raspopov: With geometry from the outer function, it's not fatal, but in my opinion it's at least bad taste, at worst may lead to problems in future.

sentinelsat/sentinel.py
valgur: Since this is a very narrowly scoped variable, the name does not matter too much. I don't think `base_path` makes too much sense as the variable name though (it's not used as a base path for anything?). Since the function parameter will not affect the variable in the outer scope and vice versa, I would keep it as it is.

sentinelsat/sentinel.py
```diff
@@ -835,19 +835,19 @@ def geojson_to_wkt(geojson_obj, feature_number=0, decimals=4):
     else:
         geometry = geojson_obj['features'][feature_number]['geometry']
 
-    def ensure_2d(geometry):
-        if isinstance(geometry[0], (list, tuple)):
-            return list(map(ensure_2d, geometry))
+    def ensure_2d(coordinates):
```
valgur: Nice, this is name fits better than the old one.

sentinelsat/sentinel.py
```diff
@@ -9,7 +9,7 @@
 import xml.etree.ElementTree as ET
 from collections import OrderedDict, defaultdict
 from contextlib import closing
-from datetime import date, datetime, timedelta
+import datetime
```
valgur: This is just a matter of taste, but I prefer to use `import datetime as dt` since it's shorter and avoids ambiguity between the module and the `datetime` object in code.

sentinelsat/sentinel.py
```diff
@@ -407,15 +407,15 @@ def get_product_odata(self, id, full=False):
...
+        url = urljoin(self.api_url, u"odata/v1/Products('{}')?$format=json".format(product_id))
         if full:
             url += '&$expand=Attributes'
         response = self.session.get(url, auth=self.session.auth)
         _check_scihub_response(response)
         values = _parse_odata_response(response.json()['d'])
         return values
 
-    def download(self, id, directory_path='.', checksum=True):
+    def download(self, product_id, directory_path='.', checksum=True):
```
valgur: `id()` is probably one of the least used builtins and shadowing it in code is not really a problem, imo. The only use for `id()` I can imagine is for super-optimized data structure or algorithm implementations, anywhere else you should be using `x is y` anyway for reference equality checks.
`product_id` is a slightly more descriptive name, I agree. I somewhat dislike modifying the public method signature, but I would not expect anyone to be using the parameter via a keyword anyway, so it should not be an issue.

sentinelsat/sentinel.py
```diff
@@ -953,15 +953,15 @@ def _format_order_by(order_by):
     output = []
     for part in order_by.split(','):
         part = part.strip()
-        dir = " asc"
+        direction = " asc"
```
valgur: 👍

sentinelsat/sentinel.py
Andrey-Raspopov: I agree that changing public interface is not the wisest move, so it should be in the changelog. In my opinion, overshadowing any built-ins can lead to the shot in the leg,  and that's why I proposed this change in the first place.

valgur:
Mostly reiterating what @j08lue already said:

- Shadowing rarely-used builtins like `id()` and `dir()` in a narrow scope is okay, IMO, and not really discouraged by PEP-8 either.
- We want to avoid breaking the signatures of public methods unless there is a strong reason to so. If you do include such changes then please definitely list them in the changelog as well.
- I actually liked the `date`->`time_interval` renaming since it makes it clearer that the variable expects an interval and also accepts timestamps, not just dates. So this something that we could include in a later release that includes other backwards-incompatible changes (such as #131).
- In case you're not doing so already - please run all unit tests locally before pushing to the PR.

sentinelsat/sentinel.py
Andrey-Raspopov: I use here the same reasoning as with geometry, that using different names in inner/outer scopes may save some trouble in future(or it may not). Anyway, right now this is definitely not a problem, so I'm not opposed to leaving it as it is.

Andrey-Raspopov:
Since some of your proposals conflict with @j08lue 's, I want to ask you to devise some commond ground, and then I'll be happy to make necessary edits.
P.S. Yeah, I've started running unit tests locally after my PR failed in the CI:)

sentinelsat/sentinel.py
j08lue: Everyone has their conventions. `dt` is for me usually a time difference, coming from physics. But perfectly fine here 😄 - so 👍 from me

sentinelsat/sentinel.py
j08lue: I feel the same about `id`, @valgur. But I think this is fine to change to get rid of any builtin-shadowing.

Fernerkundung: I would prefer to rename it to `uuid` as that is the name DHuS refers to in its [documentation](https://scihub.copernicus.eu/twiki/do/view/SciHubUserGuide/ODataAPI#Download_full_product_knowing_it).

Fernerkundung:
- disagree with `id --> product_id` and would prefer `id --> uuid`, shadowing the [DHuS docs](https://scihub.copernicus.eu/twiki/do/view/SciHubUserGuide/ODataAPI#Download_full_product_knowing_it)
- agree with `date --> time_interval`, I'm with @valgur and see that as a more descriptive name
- undecided on `import datetime as dt` as it is a matter of taste. Personally, I like it. However, [`from datetime import`](https://github.com/search?l=Python&q=from+datetime+import&type=Code) is about 9x more often used on Github than [`import datetime as dt`](https://github.com/search?l=Python&q=import+datetime+as+dt&type=Code) so I would stay with our original implementation.

j08lue:
> Personally, I like it [`dt`]

I think we should go with your likes then. 9x more popular is not a big factor. If you had said 1000x, that would be a different story.

> and would prefer `id --> uuid`

I like `uuid`, too, but it is a [core package](https://docs.python.org/3/library/uuid.html), so I think `product_id` is better.

Fernerkundung:
> I like uuid, too, but it is a core package, so I think product_id is better.

Well spotted, I forgot about the core package. +1 from me for `id` --> `product_id` then.

j08lue:
Things have diverged so much that we should start a new run at name checking, if we still deem it necessary. Sorry we let it go stale, @Andrey-Raspopov.

Issue #90, opened 2017-01-25
add support for netrc file


lucadelu:
Hi,

do you think [netrc file](https://docs.python.org/2/library/netrc.html) support could be useful? 
(I think yes so we don't need to pass user and specially password in the command line)

I started to implement it in netrc branc 52ee682

Fernerkundung:
I don't think this is a good idea, unless we can automatically determine different netrc files for different API endpoints.

For instance:
- `dhus` and `apihub` can be used with the same user:password (after an undetermined syncing period)
- the [S-3 PreOpsHub](https://scihub.copernicus.eu/s3/) uses s3guest:s3guest
- all the [national mirrors](https://github.com/Fernerkundung/awesome-sentinel#data-hubs-and-national-mirrors) require their own user:password combination

The returns of the netrc files start to diminish fast, if we want to keep `sentinelsat` as flexible as possible for all API endpoints.

What I would favour is a way to provide the password in a hashed/encrypted format, for instance when you want to use `sentinelsat` on a machine multiple people can access and you don't want your password to be plainly visible. Although I don't think that would be technically possible - unless we go down the rabbit hole of cookie/session storing?

valgur:
Both [`requests`](http://docs.python-requests.org/en/master/user/authentication/#netrc-authentication) and [`pycurl`](http://www.programcreek.com/python/example/53121/pycurl.NETRC) appear to support netrc files out of the box, surprisingly enough. It might even require relatively little additional code to get netrc support in sentinelsat, but automatically testing it is going to be painful.

Try and see if you can get the automatic support in `requests` and `homura`/`pycurl` working. Setting the auth to `None` should be sufficient for `requests`. For homura you should try adding `kwargs[pycurl.NETRC] = 1` to `api.download()`. I'm actually kind of curious to see how well this will work.

lucadelu:
@Fernerkundung yes it is possible (and already implemented) to  have different API endpoints in only one netrc file, this is mine for example

```
machine urs.earthdata.nasa.gov
login USERNASA
password PASSWDNASA

machine e4ftl01.cr.usgs.gov 
login USERNASA
password PASSWDNASA

machine https://scihub.copernicus.eu/apihub/
login USERHUB 
password PASSWDHUB
```
@valgur ok I will try in the next days

Fernerkundung:
@lucadelu I never knew - in that case it sounds like a good idea to use the netrc file if the API endpoint is included ad user:password are `None`.

[...]

PR #246, opened 2018-10-26
~/.netrc

j08lue:
This is just a stab at adding netrc support, fixing #90, as discussed there. I still think that netrc support is relevant and it is relatively easy to implement.

So here I add a little function in the CLI that tries to read `~/.netrc`, if neither command-line parameters are given nor environment variables set.

I know we could also just let `requests` handle it, probably simply by allowing no authentication to go through the API. But in case of missing auth or malformatted `~/.netrc`, that would give an error deep in the API code that is harder to communicate to the CLI user.

So I suggest to just add this in the CLI and let API users add their own `netrc` magic if they need it, since it it is very simple.

One obstacle: Testing this gets a bit dirty because I need to mess with the system environment. Also, a proper success test is only possible if I do get valid credentials from somewhere, which we currently do not have on Travis. Although maybe there is a way around the latter, just testing whether the cli fails **in the right way**...


j08lue:
The test success is a bit deceptive. Since we with VCR make requests work even if no or wrong credentials are given, the netrc test succeeds as well. But maybe that is actually OK, since this is what we do for all tests.

valgur:
👍  Maybe instead of trying to access a .netrc file in the home directory in tests we can get by with just mocking the netrc file access as is done in the [unit tests](https://github.com/requests/requests/blob/master/tests/test_requests.py#L553) of `requests` itself?

j08lue:
Good thinking, @valgur. Mocking `requests` here does not work, though, because I read the netrc file already in the CLI function.

I could follow the same logic and mokey-patch the `netrc` class here: https://github.com/python/cpython/blob/master/Lib/netrc.py#L22-L30, overriding the `__init__` function so it does not read the file. That is easy, but maybe a bit dirty. But maybe still better than my solution here and the `netrc` module will probably not change to much...? I'll do that and then you can have a look.

Do you support my choice of doing the parsing in the CLI to avoid confusion deeper down?

j08lue:
Before I do anything - do you support my choice of doing the parsing in the CLI to avoid confusion deeper down (also @Fernerkundung)? 

My concern is that if we allow omitting `user` and `password` in the SentinelAPI initialization, some users might think authorization is not required. But we could also circumvent that by making it a "power-user feature", where `user` and `password` still do not have defaults on the function but you can explicitly set them both to `None` if you know that you have a `~/.netrc` file in place. 
* Pro: That would greatly reduce the amount of extra code for this feature. 
* Con: To make this safe in the CLI, we would perhaps need to add an extra flag `--use-netrc` so omitting `-u` and `-p` without netrc still fails early.

Fernerkundung:
> do you support my choice of doing the parsing in the CLI to avoid confusion deeper down
:+1:

>Maybe we should add environment variable support to the API as well in that case (so we first try to get credentials from parameters, then envvars, then .netrc)?

In that case how would a user know whether the API takes the envvar or the .netrc file? From a design perspective I would favour it if only setting user/pass to `None` results in only one location the credentials are taken from.

Do you think we should also include a way to explicitly set the path to a `.netrc` file? A use case for that would be, if permissions on a DHuS instance are linked to user accounts and one user has to have multiple accounts, depending on the scope of the query they perform.

j08lue:
> I suggest replacing the custom `_try_netrc_auth` with `requests.sessions.get_netrc_auth(self.api_url)`

Good idea. Done. I don't think we can drop the tests, though, and they work fine as they are. So maybe we are good?

> From a design perspective I would favour it if setting user/pass to None results in only one location the credentials are taken from.

I agree. 

Actually, this already works:
```python
SentinelAPI(None, None).query(uuid='eaa093dd-46a6-49b5-a1bd-7625c2956029')
```
So all we need is to add this to the docs.

> Do you think we should also include a way to explicitly set the path to a `.netrc` file?

No, let us not do that. The `requests` parsing function does not support it and if someone has that a special use pattern, they can find other ways. Since the entries in the `.netrc` file are organized by hostname, you can have different credentials for different DHuS, e.g. a pre-ops hub etc.

So we actually support `.netrc` both in the CLI and in the API now with very little extra code. I'd consider this done and just add a few lines in the docs. Is that okay?

docs/cli.rst
```diff
@@ -15,14 +15,26 @@ password to access the Copernicus Open Access Hub.
 
   sentinelsat -u <user> -p <password> -g <search_polygon.geojson>
 
-For convenience and added security you can set the username, password and DHuS URL as environment variables and omit their entry on the command line.
+For convenience and added security, there are two ways you can store your credentials and omit them from the command line call. You can set username, password and DHuS URL as environment variables.
```
Fernerkundung: It would be great if we can include, which method takes precedence if both env-vars and a .netrc file are in place.

PR #257, opened 2019-02-06
Change default API Url

daviddemeij:
It took me some time to figure out that I should change the API URL to https://scihub.copernicus.eu/dhus. Maybe I'm misunderstanding this, but it seems to me that this should be the default URL.

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -37,7 +37,7 @@ class SentinelAPI:
         set to None to use ~/.netrc
     api_url : string, optional
         URL of the DataHub
-        defaults to 'https://scihub.copernicus.eu/apihub'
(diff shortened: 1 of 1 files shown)


j08lue:
This is a misunderstanding. The right API URL to use for automated access is the one we have implemented, `/apihub`.

The reason why that one does not work for you and `/dhus` does is that for new user accounts or those that changed passwords, there is a period of a couple of weeks or so where that change has not yet propagated to `/apihub`. This is the weirdest thing and I don't know why this is happening.

The bottom line is: Use `/dhus` until your account works with `/apihub`. I will write to SciHub support to finally find out why they have this stupid issue.

Side note: `/dhus` seems to be faster when querying sub-products (e.g. files somewhere deeper inside the .SAFE zip archive).

Issue #256, opened 2019-02-02
Missing timeout, session.post(...) blocks indefinitely


viktorbahr:
It sometimes happens to me that a host is not answering the POST request. This makes execution block forever.
https://github.com/sentinelsat/sentinelsat/blob/447e6a08b0e387f759f7c876b89ae39cf6886c6b/sentinelsat/sentinel.py#L295

`request.post` method supports `timeout` keyword. As a workaround I added `timeout` parameter to `SentinelAPI` class init. This makes `_load_subquery` raise a `requests.Timeout` exception.

Should I create a pull request?

Fernerkundung:
Sounds good, especially since `requests` proposes the inclusion of timeouts as [good practice](http://docs.python-requests.org/en/master/user/advanced/#timeouts). Good catch. 👍 

> Most requests to external servers should have a timeout attached, in case the server is not responding in a timely manner. By default, requests do not time out unless a timeout value is set explicitly. Without a timeout, your code may hang for minutes or more.

Please include a test for the timeout in your PR. This should be doable with [`requests_mock`](https://requests-mock.readthedocs.io/en/latest/response.html#raising-exceptions), which we use as part of the test suite.

PR #259, opened 2019-02-13
Add timeout to requests to avoid blocking

viktorbahr:
In reference to Issue #256 I added the `timeout` attribute to `SentinelAPI`, passing it to `_load_subquery`, `get_product_odata` and `_download` methods in order to avoid indefinite blocking in case of unresponsive DHuS.

As testing for `requests.exceptions.Timeout` was already being handled in `test_scihub_unresponsive` method I just added timeout attribute assignment and specific tests for both `requests.exceptions.ConnectTimeout` and `requests.exceptions.ReadTimeout`.


Fernerkundung:
Thank you for reporting the timeout issue **and** preparing a PR. Your changes are merged into master and I added you as a contributpor to AUTHORS.rst

PR #264, opened 2019-03-04
parsing of Online, CreationDate and IngestionDate fields

### LINKED ISSUE #233: Adding support for Long Time Archive access?
SciHub [just rolled out](https://scihub.copernicus.eu/news/News00382) a new Long Time Archive (LTA) access service. More detailed documentation is available [here](https://scihub.copernicus.eu/userguide/LongTermArchive).

The relevant details of interest to us are:
* Addition of a new OData field `Online`, which is `false` for archived products. We should definitely add this to the dictionary returned by `get_product_odata()`.
* No corresponding OpenSearch field is provided.
* Triggering of the retrieval of offline products is simply done by GET-ing the normal download URL of the product.
* Information about a retrieval being triggered and its success status is provided via [a HTML status code plus the `cause-message` header](https://scihub.copernicus.eu/userguide/LongTermArchive#HTTP_Status_codes) (which we are already using in our error messages). I'm not yet sure how to handle these, but we probably should take these into account in the code. In particular, the `202: Accepted` status code will likely cause problems for the current code, since it will not raise an exception, I think.
* There is no system to get notified about a product becoming available except for checking the OData for changes.

Adding tests for these will be slightly more difficult also, since we can't allow our code to trigger retrievals unnecessarily and their state is not constant either, so we'd need to add some mock responses.

Also, I currently do not see any LTA products being listed in SciHub: https://scihub.copernicus.eu/apihub/odata/v1/Products?$filter=Online%20eq%20false

gbaier:
This should fix #233.

I've also noticed that one of the unit tests' queries, does no longer work with the real Copernicus hub. https://scihub.copernicus.eu/apihub/odata/v1/Products('44517f66-9845-4792-a988-b5ae6e81fd3e')?$format=json reports an invalid key. Maybe this query should be updated as well.


j08lue:
Nice. So this passes all tests except for Python 3.7, which I am trying to fix in PR #266.

I don't know much about the details of this, @valgur, will you review?

And @gbaier, please add a line to the CHANGELOG, too.

j08lue:
@gbaier, build in 3.7 works again in master. Please feel free to `git cherry-pick 1bb6d7f4c7a55888f64d06e8c1940bae1ec94713` (or `git merge upstream/master`) if you want.

gbaier:
@j08lue and @valgur I've also some code for triggering the retrieval of a product from the LTA if the `online` flag is false. Although it is a little bit more invasive and probably merits a discussion how to best handle everything. Should I pile it on top of this PR, or do you want me to open a separate one?

j08lue:
> Should I pile it on top of this PR, or do you want me to open a separate one?

Very good, please do open a separate PR - maybe one of those new draft PRs. 😄

PR #267, opened 2019-03-05
Trigger retrieval from LTA if product is not online

gbaier:
Retrieval of a product from LTA is triggered if a product is not in the online archive. Obviously this PR depends on #264.

IMHO there are some points that merit a more verbose discussion:
1. What to do in case of an offline product. Right now, the code, after successfully triggering LTA retrieval, displays a warning, as proposed in https://github.com/sentinelsat/sentinelsat/issues/233#issuecomment-421242369. It does not try to download this product later.
2. How to handle errors when triggering LTA retrieval. This happens quite often since there is currently a user quota of one, concurrent retrieval in place, which causes an error. The code uses a new type of exception to pass this information along.
3. What to actually do after triggering an LTA retrieval. I suggest that, for now, users should just rerun their script after some time and hope that the product is then in the online archive. ESA just gives a time frame of 24h in which the product will become available, which in my opinion is to unspecific for automatically rescheduling a download.
4. `SentinelAPILTAError` breaks the inner loop in `download_all` to circumvent spamming the ESA server with more requests of the same product.


j08lue:
Very nice implementation @gbaier. I fully agree with your design decision to leave it to the user to try and download the products later. Breaking the download retry loop is just what needs to happen, too.

I was thinking about whether we somehow could make it more clear to the user when their `download` call triggered a request rather than actually downloading the data - apart from the warning. Also `download_all` could return a list of downloaded and a list of triggered LTA products so the user has full control over what to do with the triggered products.

But maybe we should leave that for later when we get some feedback on how useful the API is and whether someone misses any such feature.

So I think this is actually ready for merge.

sentinelsat/__init__.py
```diff
@@ -3,5 +3,5 @@
 # Import for backwards-compatibility
 from . import sentinel
 
-from .sentinel import SentinelAPI, SentinelAPIError, InvalidChecksumError, read_geojson,\
-    geojson_to_wkt, format_query_date
+from .sentinel import InvalidChecksumError, SentinelAPI, SentinelAPIError, SentinelAPILTAError, \
```
j08lue: Should we maybe start using `__all__` in `sentinel.py` to select the imports?

sentinelsat/sentinel.py
```diff
@@ -428,6 +428,39 @@ def get_product_odata(self, id, full=False):
...
+        ----------
+        url : string
+            URL for downloading the product
+
+        Notes
+        -----
+        https://scihub.copernicus.eu/userguide/LongTermArchive
+
+        """
+        # wrapping in closing no longer necessary https://github.com/kennethreitz/requests/pull/4137
```
j08lue: Nice and thanks for the note about `closing`. Please remove the comment though, now that `closing` is gone.

tests/test_mod.py
```diff
@@ -725,6 +735,44 @@ def test_scihub_unresponsive():
             api.download_all(['8df46c9e-a20c-43db-a19a-4240c2ed3b8b'])
 
 
+def test_trigger_lta():
```
j08lue: Maybe split this up into several tests for more granular checks? Like `test_trigger_lta_accepted`, `test_trigger_lta_fails_not_accepted`, ...

tests/test_mod.py
```diff
@@ -725,6 +735,44 @@ def test_scihub_unresponsive():
...
+
+    with requests_mock.mock() as rqst:
+        rqst.get(
+            request_url,
+            text="Mock trigger downloading an offline file", status_code=500
+        )
+        with pytest.raises(SentinelAPILTAError) as excinfo:
+            api._trigger_offline_retrieval(request_url)
+
+
```
j08lue: Please add a higher-level test, too, that calls `api.download` to get the `break` catch covered.

CHANGELOG.rst
```diff
@@ -17,6 +17,7 @@ Added
 * Added support for Unicode symbols in search queries. (#230)
 * Raise ValueError exception if longitude is outside [-180, 180] or latitude is outside [-90, 90] (#236, #218 @Andrey-Raspopov)
 * optional ``timeout`` attribute to avoid indefinite wait on response from the server (#256, @viktorbahr)
+* Parsing the ``Online``, ``CreationDate`` and ``IngestionDate`` fields of an OData response
```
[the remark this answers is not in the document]
gbaier: Done

sentinelsat/sentinel.py
gbaier: Done. This was mostly a note that it could be removed in other places as well.

sentinelsat/__init__.py
gbaier: While I agree, maybe we should leave this for a separate edit/PR. It might get too confusing.

tests/test_mod.py
gbaier: Also done. Do you think it makes sense to provide a fixture for these very similar unit tests?

tests/test_mod.py
gbaier: I've added a new test for `download_all` with a product in the LTA.

gbaier:
> I was thinking about whether we somehow could make it more clear to the user when their `download` call triggered a request rather than actually downloading the data - apart from the warning. Also `download_all` could return a list of downloaded and a list of triggered LTA products so the user has full control over what to do with the triggered products.

I've added this functionality to `download_all` . The somewhat over-verbose code with the ordered dictionaries and list comprehensions should preserve the order of the processed products. Obviously the additional return value might break existing user code.

gbaier:
Any news of further comments about this PR?

I've also written some initial code that waits for products from the LTA to become available, while downloading products that are already online. But since it uses concurrent.futures it can't run on stock Python 2.7. Let me know if there is any interest.

tests/test_mod.py
j08lue: Yes, good idea. For the `lta_url` and the `api_mock` (or `mock_api`?).

j08lue:
Thanks for the ping, @gbaier. Sorry for the silence. @valgur, would you have time to review this? Or you @Fernerkundung? I have not yet used LTA data, so I can't judge the UX too well.

> I've also written some initial code that waits for products from the LTA to become available, while downloading products that are already online.

If not else, that recipe would be great to have in the docs! (Regarding Python 2, I think we could soon ditch it. But that is a different issue...)

Fernerkundung:
I currently don't have time to do a full review on this. From a quick glance however everything looks great.

@j08lue I wouldn't worry about the UX too much right now. In the past we have seen the hub to issue multiple smaller changes after a major change. I'm certain they will adapt the DHuS behaviour regarding LTA and then we have to refactor the code anyway.

Fernerkundung:
Regarding the concurrent download of LTA retrieved products and online products:

That would be a great addition to the docs/examples @gbaier :+1: 

Including that logic into the API would mean dropping 2.7 support and refactoring the check for concurrent download streams (Copernicus Open Access Hub only allows 2 concurrent downloads, most other hubs have no limitation). My vote would be to wait with the inclusion into the API until we see the specifics of the LTA hashed out (from the Copernicus DHuS side) and how the LTA retrieval and performance evolves.

I also don't thinkwe should drop 2.7 support too soon. The main users of Copernicus data are scientists and government agencies- both of which tend to have a large number of legacy systems that are still on 2.7 (and probably will be for some time).

j08lue:
Thanks @Fernerkundung, let us merge this PR then (maybe after the tiny [addition of fixtures for the tests](https://github.com/sentinelsat/sentinelsat/pull/267#discussion_r265141871)). 

We need to update the docs, though (under Python API > LTA products). Would you have time to start another PR for that, @gbaier? I am happy to contribute. We can include your Python 3 code for automatic download there.

tests/test_mod.py
gbaier: I could not think of an elegant way with a fixture. The unit tests for unsuccessful LTA requests are now parameterized with `pytest.mark.parametrize`.

gbaier:
I did the requested cosmetic change for [making the unit tests more concise](https://github.com/sentinelsat/sentinelsat/pull/267#issuecomment-472787102).

Regarding the code for concurrent downloading and requesting from LTA. My current implementation is a complete rewrite of `download_all`. Due to Python 2.7 support and since the whole procedure for retrieving products from the LTA is still being phased-in and their might be some changes, I also wouln't add the code anytime soon.

@j08lue I've added a PR for the LTA documentation: #268

tests/test_mod.py
j08lue: nice, thanks!

j08lue:
Great work, @gbaier, thanks a lot! I am merging this now.

Issue #281, opened 2019-04-15
No 'Online' key in Collaborative Ground Segment OData response breaks _parse_odata_response


viktorbahr:
I noticed that some Collaborative Ground Segments (i.e. https://collaborative.asi.it) do not return an `Online` key in their OData response.

Since #264 this key is required, otherwise [`_parse_odata_response`](https://github.com/sentinelsat/sentinelsat/blob/6a23ebfd53ca80634fb691fb9b887dd37964f934/sentinelsat/sentinel.py#L1128) will raise a `KeyError`.

This effectively breaks OData interaction with respective mirrors (`get_product_odata`, `download` etc).

After briefly looking over it, I would assume that a simple

```python
if 'Online' in product:
```

in `_parse_odata_response` could fix this.

Any thoughts on this? Should I create a PR?

valgur:
Should replace `product['Online']` with `product.get('Online')` so it returns `None` if missing.
A PR is always welcome, sure.

viktorbahr:
Okay, will submit a PR later.

viktorbahr:
I'd like to create a test for this, too.
Can anyone point out a clever way to do this? My problem is:
- Only some collaborative ground segments return no `Online` key in OData response (ESA SentinelHub should __always__ contain it)
- So, recording a response from ESA SentineHub is out of the question, correct?
- Should I instead record a new casette using a different mirror (i.e. add an additional `api_kwargs` with different `api_url` that I'll only use for my `test_product_odata_short_with_missing_online_key` test)?
- The latter approach would require registering an account at this mirror using the `DHUS_USER`, `DHUS_PASSWORD` variables stored at Travis CI, right?
- Retention time for products at Collaborative Ground Segment mirrors is significantly shorter than at ESA SentinelHub. What happens if the product is dropped? Do we just pick a different product?

j08lue:
Maybe you could also monkey-patch a SciHub response. Removing that one key should be simple enough. Maybe VCR.py even has a function for that, so it is easily repeatable. Too dirty?

viktorbahr:
It was quite easy to remove the key-value pair from a prerecorded response using pyVCRs `before_record_response` hook.
Thanks for pointing me in this direction!

Fernerkundung:
Closed after merging #282

PR #282, opened 2019-04-16
Handling non-existant 'Online' key in OData response

viktorbahr:
In reference to Issue #281 I am using dicts `get` method to avoid `KeyError` on missing `Online` key.

I also added a test for this that simply scrubs the key-value pair from a prerecorded response using pyVCRs `before_record_response` hook.

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -1125,7 +1125,7 @@ def _parse_odata_response(product):
         'date': _parse_odata_timestamp(product['ContentDate']['Start']),
         'footprint': _parse_gml_footprint(product["ContentGeometry"]),
         'url': product['__metadata']['media_src'],
-        'Online': product.get('Online'),
+        'Online': product.get('Online', True),
         'Creation Date': _parse_odata_timestamp(product['CreationDate']),
         'Ingestion Date': _parse_odata_timestamp(product['IngestionDate']),
     }
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -1125,7 +1125,7 @@ def _parse_odata_response(product):
         'date': _parse_odata_timestamp(product['ContentDate']['Start']),
         'footprint': _parse_gml_footprint(product["ContentGeometry"]),
         'url': product['__metadata']['media_src'],
-        'Online': product['Online'],
+        'Online': product.get('Online'),
         'Creation Date': _parse_odata_timestamp(product['CreationDate']),
         'Ingestion Date': _parse_odata_timestamp(product['IngestionDate']),
     }
--- a/tests/test_mod.py
+++ b/tests/test_mod.py
@@ -501,6 +501,41 @@ def test_get_product_odata_short():
         for k in ret:
             assert ret[k] == expected[k]
 
+def scrub_string(string, replacement=''):
+    '''Scrub a string from a VCR response body string
+    '''
+    def before_record_response(response):
+        response['body']['string'] = response['body']['string'].replace(string, replacement)
+        return response
+    return before_record_response
+
+@pytest.mark.scihub
+def test_get_product_odata_short_with_missing_online_key():
+    api = SentinelAPI(**_api_auth)
+
+    uuid = '8df46c9e-a20c-43db-a19a-4240c2ed3b8b'
+    expected_short = {
+        'id': '8df46c9e-a20c-43db-a19a-4240c2ed3b8b',
+        'size': 143549851,
+        'md5': 'D5E4DF5C38C6E97BF7E7BD540AB21C05',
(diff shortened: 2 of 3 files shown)


sentinelsat/sentinel.py
```diff
@@ -1125,7 +1125,7 @@ def _parse_odata_response(product):
         'date': _parse_odata_timestamp(product['ContentDate']['Start']),
         'footprint': _parse_gml_footprint(product["ContentGeometry"]),
         'url': product['__metadata']['media_src'],
-        'Online': product['Online'],
+        'Online': product.get('Online'),
```
j08lue: I wonder what `'Online'` should default to. Now it defaults to `None`. Maybe `True` would be better? Because when you pass the response with `'Online': None` to `api.download`, it will go into the LTA access triggering branch and that will be bad...

So how about `product.get('Online', True)`?

tests/test_mod.py
```diff
@@ -501,6 +501,41 @@ def test_get_product_odata_short():
...
+        'footprint': 'POLYGON((-63.852531 -5.880887,-67.495872 -5.075419,-67.066071 -3.084356,-63.430576 -3.880541,'
+                     '-63.852531 -5.880887))',
+        'title': 'S1A_EW_GRDM_1SDV_20151121T100356_20151121T100429_008701_00C622_A0EC',
+        'Online': None,
+        'Creation Date': datetime(2015, 11, 21, 13, 22, 1, 652000),
+        'Ingestion Date': datetime(2015, 11, 21, 13, 22, 4, 992000),
+    }
+
+    # scrub 'Online' key from response
+    with my_vcr.use_cassette("test_get_product_odata_short.yaml",
```
j08lue: nice

tests/test_mod.py
```diff
@@ -617,7 +652,6 @@ def test_get_product_odata_full():
         for k in ret:
             assert ret[k] == expected[k]
 
-
```
j08lue: shouldn't there actually be two blank lines here? :-P

sentinelsat/sentinel.py
viktorbahr: Good call, will default to `True`.

tests/test_mod.py
viktorbahr: 🙈

j08lue:
Cool. Now I think the only thing left is to add a line to the `CHANGELOG`. Will you do that, @viktorbahr?

viktorbahr:
Yes, will do.

tests/test_mod.py
```diff
@@ -502,6 +502,42 @@ def test_get_product_odata_short():
             assert ret[k] == expected[k]
 
 
+def scrub_string(string, replacement=''):
```
Fernerkundung: Is there any reason to nest function definitions here? I would propose to change this to one function.

tests/test_mod.py
```diff
@@ -502,6 +502,42 @@ def test_get_product_odata_short():
...
+                     '-63.852531 -5.880887))',
+        'title': 'S1A_EW_GRDM_1SDV_20151121T100356_20151121T100429_008701_00C622_A0EC',
+        'Online': True,
+        'Creation Date': datetime(2015, 11, 21, 13, 22, 1, 652000),
+        'Ingestion Date': datetime(2015, 11, 21, 13, 22, 4, 992000),
+    }
+
+    # scrub 'Online' key from response
+    with my_vcr.use_cassette("test_get_product_odata_short.yaml",
+                             before_record_response=scrub_string(b'"Online":false,', b'')):
```
Fernerkundung: This should be `False` to be consistent with PEP285.

Fernerkundung:
Good catch and thank your for preparing this PR. I just requested two minor changes, then this can be merged.

tests/test_mod.py
viktorbahr: Function definition nesting is used here to enable passing of `scrub_string` method as `before_record_response` callback. By default, callbacks need to be of format `callback_name(request)` (inferred from [vcrpy docs](https://vcrpy.readthedocs.io/en/latest/advanced.html#custom-response-filtering)). By nesting the `before_record_response` method the `scrub_string` methods allows scrubbing of _any_ string, not just hardcoded ones.

Could change this to a more specific `scrub_online_kv` callback?

tests/test_mod.py
viktorbahr: IMO PEP285 does not apply here because the the string passed to the `scrub_string` method is not referring to a Python variable but a value inside the [`test_get_product_odata_short.yaml`](https://github.com/viktorbahr/sentinelsat/blob/d3046e8ec385a23f9d04d070adccd0a620f48abc/tests/fixtures/vcr_cassettes/test_get_product_odata_short.yaml#L12) fixture.

Changing to uppercase `b'"Online":False,'` would not match the target value in the respective fixture, wouldn't it?

Fernerkundung: You are absolutely right. My bad - I didn't catch that this was taken verbatim from the fixture.

Issue #247, opened 2018-11-02
Improve VCR recording - do not record bad responses?


j08lue:
Thanks to @anujkumar93 for pointing out a way to delete bad vcrpy recordings (https://github.com/sentinelsat/sentinelsat/pull/86#issuecomment-435121322).

This issue is just to formally raise the question whether we should implement something like this - or not for now.

I have not experienced any problems with recording bad responses. But I have not been using vcrpy too much, just occasionally when writing new tests and then there were no problems. So from my experience, this is not needed right now.

What do you think, @valgur and @Fernerkundung?

Fernerkundung:
I haven't had any problems with failed tests as well as my workflow was usually to ensure all tests pass and then record new cassettes.

Maybe we wait until kevin1024/vcrpy#205 is implemented?

If it is more of an issue than I have encountered I also don't see a lot of harm in including the proposed patch (i.e. it does not entail a lot of code complexity and technical debt).

valgur:
From my experience, I'd say that the recordings are still very useful even when the tests fail. The query and response content usually stay the same even after fixing the test, but the cassettes speed up the test runs. I delete and re-record the affected cassette before committing. #453 also adds `clean_cassettes.sh` and speed ups the live test runs, so re-recording all cassettes is also a fairly convenient option now.

PR #283, opened 2019-04-22
Unit test refactoring, update cassettes

valgur:
Replaced direct `vcrpy` usage in unit tests with `pytest-vcr`, which provides almost identical functionality in a cleaner manner, and replaced quite a bit of boilerplate code in unit tests with equivalent pytest fixtures. Notably, `test_cli.py` became 25% shorter as a result.

Related changes:
* The `pytest` command line options changed from `--vcr disable` to `--disable-vcr` and
  `--vcr [use|record_new|reset]` to `--vcr-record [once|record_new|all]`.
* Merged `shared.py` into `conftest.py`.
* Reduced code duplication in unit tests by making heavy use of pytest fixtures.
* Moved some of the more verbose test data into separate fixture files.
* Replaced the context managers for netrc tests in test_cli.py with pytest fixtures.

I would expect #247 to be significantly easier to implement as a result of using `pytest-vcr` as well.


valgur:
The updated cassette files make the diff view of the PR's changes very noisy, so I suggest hiding them: https://github.com/sentinelsat/sentinelsat/pull/283/files?file-filters%5B%5D=.py&file-filters%5B%5D=.rst

docs/install.rst
```diff
@@ -29,8 +29,8 @@ To allow the tests to run actual queries against Copernicus Open Access Hub set
     export DHUS_USER=<your scihub username>
     export DHUS_PASSWORD=<your scihub password>
 
-and add ``--vcr disable`` to ``py.test`` arguments.
-To update the recordings use either ``--vcr record_new`` or ``--vcr reset``.
+and add ``--disable-vcr`` to ``py.test`` arguments.
```
Fernerkundung: should also be ``pytest`` for consistency

valgur: fixed, thanks!

PR #293, opened 2019-06-07
Offline products retrieval from LTA

gbaier:
This is some initial code for **download_all** to handle offline products. **download_all** now

1. starts downloading online products right away
2. requests the retrieval of offline products from the LTA (There is a limit of 1 request every 30minutes per user)
3. waits for offline products to become available and then starts downloading them (There are a couple of retries and timeouts)

I consider the code to be only a draft, with some points still needing to be discussed. What essentially needs to be done is downloading products while requesting offline products and waiting for them to become available.

This can be either implemented using **concurrent.futures** or **asyncio**.  **concurrent.futures** is in my opinion less elegant but should be Python2 compatible if users install the corresponding pip packages (https://pypi.org/project/futures/). **asyncio** breaks Python2, which personally I don't care about, but in that case we should probably also replace **requests** with **aiohttp**, which would be a bigger overall change.

Next question in my opinion is how to wait for offline products. Indefinitely? As long as downloading online products? Or with a timeout? I'd love to hear your opinion on that.

In case I missed something please let me know. I can merge previous commits, squash my commits and in general clean up my PR to make it ready for merging when we have decided on how to proceed.


Fernerkundung:
> Next question in my opinion is how to wait for offline products. Indefinitely? As long as downloading online products? Or with a timeout? I'd love to hear your opinion on that

I vote for waiting as long as downloading online products and providing a message to the user that LTA retrieval was triggered and they should retry the query after the retrieval was done. My reasoning is that there are a lot of tools out there that enable the recurring retry of the download based on the LTA retrieval or even recurring times (workflow tools, cron jobs, etc.). I say we stick with the Unix philosophy "Make each program do one thing well" instead of expanding the functionality of `sentinelsat` into something that can be handled by other tools, which also provide the user with more flexibility how they use `sentinelsat`. It won't hurt to include an example how to do this in the documentation, though.

> Python2 compatible if users install the corresponding pip packages

Adding this to the requirements.txt is not a big issue. The discussion would then be, how long do we aim for keeping up Python2 support? EOL for Python 2 is end of this year. A number of our users are probably in environments where they can not upgrade easily. Therefore I would tend towards using `concurrent.futures` unless we find a compelling argument for `asyncio`.

gbaier:
> I vote for waiting as long as downloading online products and providing a message to the user that LTA retrieval was triggered and they should retry the query after the retrieval was done. My reasoning is that there are a lot of tools out there that enable the recurring retry of the download based on the LTA retrieval or even recurring times (workflow tools, cron jobs, etc.). I say we stick with the Unix philosophy "Make each program do one thing well" instead of expanding the functionality of `sentinelsat` into something that can be handled by other tools, which also provide the user with more flexibility how they use `sentinelsat`.

Just to make sure that I understand correctly. You mean that **download_all** would still in the background trigger the retrieval of products from the LTA, but stops as soon as there is nothing to download? I agree, that otherwise things would get too complicated.

Only two more questions / proposals.
1) In addition to `max_attempts` I would add `retry_delay` as another parameter to **download_all**, which would wait this time in seconds between failed download attempts. If a product is not yet online this would also count as a failed download. This way users could IMHO easily tune  **download_all** to wait for offline products.
2) What to do if all products that are to be downloaded are offline. Trigger retrieval of the first and then terminate? Or just put it on the download queue and try `max_attempts`

> Therefore I would tend towards using `concurrent.futures` unless we find a compelling argument for `asyncio`.

I agree and for our use case I don't think asyncio brings much to the table. The APIs are somewhat similar so switching at a later stage is not that big of a deal.

Fernerkundung:
>What to do if all products that are to be downloaded are offline. Trigger retrieval of the first and then terminate? Or just put it on the download queue and try `max_attempts`

I would say the expected behaviour is to trigger as many as possible and terminate if no more products can be downloaded. Then inform the user of the number of downloads, triggered products, untriggered products and that they should retry after the LTA retrieval has been finished. This way the behaviour would be identical, whether the number of downloaded products is 0 or >0. Attempting to download and triggering if the product is in LTA is something I would regard as "download" - therefore `sentinelsat` should continue to "download", unless nothing can be downloaded/triggered anymore.

>`retry_delay` as another parameter to download_all, which would wait this time in seconds between failed download attempts.

This is a -1 from my side. There are a number of libraries that implement delayed retries, also supporting async functions. Instead of including them into our code I think an example in the documentation would be enough. 

From Python this could be done with the [tenacity library](https://github.com/jd/tenacity).

```python
from tenacity import retry, wait_fixed, stop_after_attempt

@retry(wait=wait_fixed(1800) | stop_after_attempt(5))
api.download_all(products)
````

In a linux environment one could use `parallel` for delayed retries.

```bash
parallel --jobs 1 --retries 5 --delay 30m ::: sentinelsat -d ...
```

j08lue:
> I say we stick with the Unix philosophy "Make each program do one thing well" instead of expanding the functionality of `sentinelsat` into something that can be handled by other tools, which also provide the user with more flexibility how they use `sentinelsat`.

In principle, I agree that we should not turn `download_all` into a full workflow manager that async waits on triggered products and (re-)tries downloading them later. On the other hand, exactly this - handling the delay and all the bookkeeping involved - is hard to get right, so a good implementation here would enable many people to continue using SciHub for offline products who are otherwise pretty lost.

As for Python 2 support: I am really in favor of dropping that very soon - as soon as we would have to go lengths for compatibility. If that is the case here, maybe it is time. Let's evaluate that when/if we get to the concurrency part.

Fernerkundung:
>  a good implementation here would enable many people

That is a convincing argument.
So, as a middle ground: implement a lightweight dependency (i.e. tenacity which is pure Python) and keep the default to not wait and retry?

gbaier:
> So, as a middle ground: implement a lightweight dependency (i.e. tenacity which is pure Python) and keep the default to not wait and retry?

Sounds good to me. I'll first adjust my implementation for concurrent downloading and triggering from the LTA without retrying. As a second step then retrying with tenacity on top of `download_all`. I have to admit that I was not aware of tenacity, but it seems to be a nice and lightweight solution.

gbaier:
Since I've cleaned up the code and also merged other commits on another branch I'll close this PR draft to not make it too messy.

PR #297, opened 2019-06-20
download and retrieve offline products from LTA

gbaier:
This PR is a follow-up to #293 and is mostly a complete rewrite of `download_all` to handle offline products. `download_all` downloads all online products while trying to periodically retrieve offline products from the LTA. As soon as there are no longer any products to download it terminates.

Things to note:
1. I had slight problems with unit tests. `smallest_online_products.yaml` can change from call to call if an
    offline product becomes available. Also dhus and apihub have different md5 hashes for essentially the same products. I managed to get all unit tests to pass but there might now be some redundancies.
2. I tried to only modify `download_all` and leave everything else alone. If we decide to also change `download` there shouldn't be any need for the wrapper function `_download_online_retry`. This will however break existing code since `download` would need to act differently for offline products.
3. As suggested by @Fernerkundung we could wrap `download_all` with the [tenacity library](https://github.com/jd/tenacity) to wait longer for offline products. Adding tenacity as a dependency would also allow us to write a nicer `_trigger_offline_retrieval_until_stop`. I'm still somewhat opposed to tenacity, mainly because its another dependency. But IMHO it definitely makes sense to add an example to the documentation how users could use it. Also, while admittedly my implementation of  `_trigger_offline_retrieval_until_stop` is not too elegant it with threading and everything, I think it is simple enough not to cause any problems.
4. I only tested with Python 3.6. Python 2.X needs https://pypi.org/project/futures/ as a depedency.

I hope this meets your expectations. Any comments and suggestions are welcome.


Fernerkundung:
Thank you for contributing and sorry for the delay on your PR @gbaier. I currently do not have the time to review it. Maybe @j08lue or @valgur could step in?

gbaier:
**Bump**. Any comments or suggestions?

Fernerkundung:
If no one else is able to step in I should be able to find some time beginning of next week.

sentinelsat/sentinel.py
```diff
@@ -587,44 +614,168 @@ def download_all(self, products, directory_path='.', max_attempts=10, checksum=T
...
-        set[string]
-            The list of products that failed to download.
+            A dictionary containing the product information for products successfully
+            triggered for retrieval from the long term archive but not downloaded.
+        dict[string, dict]
+            A dictionary containing the product information of products where either
+            downloading or triggering failed
         """
+
+        lta_retry_delay = 600 # try to request a new product from the LTA every 600 seconds
```
Fernerkundung: Do we need to expose this as a function parameter or class attribute?

tests/test_cli.py
```diff
@@ -3,13 +3,14 @@
 import re
 import shutil
 from contextlib import contextmanager
+from functools import partialmethod
```
Fernerkundung: This is not supported in Python 2.7, where it would be `from functools import partial`

Fernerkundung:
> I had slight problems with unit tests. smallest_online_products.yaml can change from call to call if an offline product becomes available.

Good to note. Using the recorded responses this should not be problem for our unit tests. Might be good to put this as a comment in the unit test so people can find it if they re-record the responses.

> But IMHO it definitely makes sense to add an example to the documentation how users could use it.

Your solution looks maintaineable enough to warrant the argument of not including an additional library. If you could include an example into the docs that would be great.

Please also document the LTA trigger behaviour and the new options for concurrent trigger/download.

sentinelsat/sentinel.py
```diff
@@ -428,37 +430,27 @@ def get_product_odata(self, id, full=False):
         values = _parse_odata_response(response.json()['d'])
         return values
 
-    def _trigger_offline_retrieval(self, url):
-        """ Triggers retrieval of an offline product
 
-        Trying to download an offline product triggers its retrieval from the long term archive.
-        The returned HTTP status code conveys whether this was successful.
+    def is_online(self, id):
```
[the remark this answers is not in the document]
gbaier: I added unit tests. The code can now also handle non-existent products.

sentinelsat/sentinel.py
```diff
@@ -428,37 +430,27 @@ def get_product_odata(self, id, full=False):
...
-                raise SentinelAPILTAError('Requests for retrieval from LTA exceed user quota', r)
-            elif r.status_code == 500:
-                # should not happen
-                self.logger.error("Trying to download an offline product")
-                raise SentinelAPILTAError('Trying to download an offline product', r)
-            return r.status_code
+            if r.status_code == 200:
+                return r.text == 'true'
+            else:
+                raise SentinelAPIError('Could not verify whether product {} is online'.format(id), r)
```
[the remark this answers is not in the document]
gbaier: Fixed

sentinelsat/sentinel.py
gbaier: Passed as a function parameter

tests/test_cli.py
gbaier: I added some boilerplate code to handle the import in Python 2.7.

gbaier:
> > I had slight problems with unit tests. smallest_online_products.yaml can change from call to call if an offline product becomes available.
> 
> Good to note. Using the recorded responses this should not be problem for our unit tests. Might be good to put this as a comment in the unit test so people can find it if they re-record the responses.
> 
Done.

> > But IMHO it definitely makes sense to add an example to the documentation how users could use it.
> 
> Your solution looks maintaineable enough to warrant the argument of not including an additional library. If you could include an example into the docs that would be great.

I added a short example to the LTA product section

> Please also document the LTA trigger behaviour and the new options for concurrent trigger/download.

IIRC the docstrings already mention all options, or did I miss something?

> > Python 2.X needs https://pypi.org/project/futures/ as a depedency.
> 
> I added the futures dependency, but there are still some minor functools issues for the 2.7 unit tests to pass.

Thanks. I added a workaround for Python 2.X.

Fernerkundung:
Thank you for this extensive PR _and_ your patience @gbaier! :confetti_ball:

PR #305, opened 2019-07-31
Parallel functioanlity 

ziya5635:
Parallel functionality added, in addition, some extra error handling add to prevent crashing when there are serverside issues, which is the current problem of Fin-hub servers. Please let me know if you have some comments on these changes.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="JavaScriptSettings">
+    <option name="languageLevel" value="ES6" />
+  </component>
+  <component name="ProjectRootManager" version="2" project-jdk-name="Python 3.7 (gis) (2)" project-jdk-type="Python SDK" />
+</project>
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -23,6 +23,8 @@ from tqdm import tqdm
 
 from . import __version__ as sentinelsat_version
 
+from multiprocessing.pool import ThreadPool
+from multiprocessing import Queue
 
 class SentinelAPI:
     """Class to connect to Copernicus Open Access Hub, search and download imagery.
@@ -73,6 +75,11 @@ class SentinelAPI:
         # For unit tests
         self._last_query = None
         self._last_response = None
+        # keep track of failed downloads
+        self.failed_list_id = list()
+        self.failed_list_title = list()
+        # Keep track of the products downloaded successfully
+        self.successful_list_title = list()
 
     def query(self, area=None, date=None, raw=None, area_relation='Intersects',
               order_by=None, limit=None, offset=0, **keywords):
@@ -499,7 +506,9 @@ class SentinelAPI:
 
         if exists(path):
             # We assume that the product has been downloaded and is complete
+            self.successful_list_title.append(product_info['title'])
             return product_info
+
 
         # An incomplete download triggers the retrieval from the LTA if the product is not online
         if not product_info['Online']:
@@ -539,16 +548,29 @@ class SentinelAPI:
         if not skip_download:
             # Store the number of downloaded bytes for unit tests
             product_info['downloaded_bytes'] = self._download(
-                product_info['url'], temp_path, self.session, product_info['size'])
+                product_info['url'], temp_path, self.session, product_info['size'], product_info=product_info)
 
         # Check integrity with MD5 checksum
         if checksum is True:
-            if not self._md5_compare(temp_path, product_info['md5']):
-                remove(temp_path)
-                raise InvalidChecksumError('File corrupt: checksums do not match')
+            try:
+                if not self._md5_compare(temp_path, product_info['md5']):
+                    remove(temp_path)
+                    if product_info['title'] not in self.failed_list_title:
+                        self.failed_list_title.append(product_info['title'])
+                        self.failed_list_id.append(product_info['id'])
+                    raise InvalidChecksumError('File corrupt: checksums do not match')
+            except Exception as e:
+                # prevents crashing when internal server error occurs, causing no files are created for checking sum
+                self.logger.critical(e)
+
+        # keep track of failed downloads
+        if skip_download is True and product_info['id'] not in self.failed_list_id:
+            self.failed_list_id.append(product_info['id'])
+            self.failed_list_title.append(product_info['title'])
 
         # Download successful, rename the temporary file to its proper name
-        shutil.move(temp_path, path)
+        if len(self.successful_list_title) > 0:
+            shutil.move(temp_path, path)
         return product_info
 
     def download_all(self, products, directory_path='.', max_attempts=10, checksum=True):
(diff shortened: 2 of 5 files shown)


ziya5635:
Hi, Travis says download single and download all for some reasons have failed. It is working well on my local!

ziya5635:
Ok, I found an issue to solve in my code. I will fix it shortly and make a new request.

j08lue:
No need to close and create new pull requests. Just add to your branch. Also, consider making it a draft pull request instead.

ziya5635:
ok, that's a better idea. thanks!

j08lue:
Thanks for submitting a PR, @ziya5635.

Before you continue working on it - I am pretty sure parallel download is beyond the scope of the package. We usually leave this kind of boilerplate code / workflow management up to the user to implement and keep our implementation to a minimum that is required to interact with the logic of a DHuS. What do you think, @Fernerkundung and @valgur?

For potential future PRs, please make sure not to add and commit editor files like `.idea`, like you are doing here. And try not to change the current API, unless you cannot avoid it. Perhaps start with smaller issues where we can help you curate your code so it fits with our style choices (e.g. no blank `try ... except` etc.).

Fernerkundung:
>  parallel download is beyond the scope of the package. We usually leave this kind of boilerplate code / workflow management up to the user to implement and keep our implementation to a minimum that is required to interact with the logic of a DHuS.

:+1: fully agree to this.

What is the issue you are trying to adress here @ziya5635? Does this significantly improve download speeds for you?

We could introduce a parallel function to `download_all` that uses concurrent downloads. However, that would soon let us run into rather complicated design decisions, e.g. threading or parallelism, async, sensible defaults?

Adding to this, concurrent downloads usually do not provide a increase in download speed to the user. This is due to a number of reasons:

- most DHuS do not rate limit single connections but rather limit the number of connections
- the files are large, therefore the overhead is minimal and multiple threads do not achieve significant improvements
- for most connections I tested (up to VMs in probably the same data center as SciHub) a single download will saturate the transfer speed of the connection

ziya5635:
I started communicating about this functionality with Valgur about 2 months ago and I think he shared and asked the other members about this. The response I got was "implement the work and make a pull request to us, after approval it would be merged". Regarding your wonder about speed, yes it does improve the time takes to download products. Probably you are judging based on ESA servers, which provides limited service to users. This functionality shows itself when the server offers you a wider band and you have a good internet connection. Servers such as Fin-Hub provide unlimited downloaded at a time with a much wider band. I have been downloading and using this tool and it has indeed saved a lot of my time when one is interested to do time series of analysis, which usually requires downloading a number of products. I encourage you to test this functionality using other servers that provide unlimited services to users and feel the difference.

Fernerkundung:
Please do not consider the comments posted by @j08lue and me as disapproval but rather discussion. This project does not have leaders per-se but rather we try to evaluate changes to the packages by their merit and complexity. Part of this is, that major changes to the package are discussed in the open - enabling anyone wo would like to add their comment.

As @valgur responded to your Email:
> Contributions to sentinelsat are always appreciated and both of the suggested improvements sound useful, indeed. We have an established and well-working procedure in place for handling code contributions, though, which is via pull requests that are reviewed and approved.

Please regard the discussions happening in here as part of the review process.

> I encourage you to test this functionality using other servers that provide unlimited services to users and feel the difference.

That would be a very good starting point to evaluate the merit of concurrent downloads. Can you create a small self sustained example script (minus login credentials) that allows for the testing and comparison of single vs. parallel downloads from SciHub and FinHub? As part of the discussion I'd like to run this on some machines as well. What I posted above are just assumptions - an example would enable a discussion based on actual data, which is preferable.

If you want to go forward with this PR please also be prepared for code-style review. For instance the comments already posted by @j08lue.

ziya5635:
Yes, of course, I let you know when it is ready.

ziya5635:
Hi, I wonder if you have come to any conclusion about parallel functionality? have you hopefully manage to test the code? Should I continue working on the code or stop it? Please let me know about your thoughts on this.

.idea/misc.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
j08lue: This is editor stuff. Should not be committed. Please remove.

sentinelsat/sentinel.py
```diff
@@ -73,6 +75,11 @@ def __init__(self, user, password, api_url='https://scihub.copernicus.eu/apihub/
         # For unit tests
         self._last_query = None
         self._last_response = None
+        # keep track of failed downloads
+        self.failed_list_id = list()
```
j08lue: You add a lot of state to the API here, that is, some lists that will keep growing over subsequent usage of the instance. Better to leave the tracking of failed downloads to the user.

sentinelsat/sentinel.py
```diff
@@ -539,16 +548,29 @@ def download(self, id, directory_path='.', checksum=True):
...
-                remove(temp_path)
-                raise InvalidChecksumError('File corrupt: checksums do not match')
+            try:
+                if not self._md5_compare(temp_path, product_info['md5']):
+                    remove(temp_path)
+                    if product_info['title'] not in self.failed_list_title:
+                        self.failed_list_title.append(product_info['title'])
+                        self.failed_list_id.append(product_info['id'])
+                    raise InvalidChecksumError('File corrupt: checksums do not match')
+            except Exception as e:
```
j08lue: Never do this. Catchall and log is really bad practice because it obscures errors. You need to log and skip failing downloads somehow else.

sentinelsat/sentinel.py
```diff
@@ -806,30 +885,43 @@ def _md5_compare(self, file_path, checksum, block_size=2 ** 13):
...
-            mode = 'ab' if continuing else 'wb'
-            with open(path, mode) as f:
-                for chunk in r.iter_content(chunk_size=chunk_size):
-                    if chunk:  # filter out keep-alive new chunks
-                        f.write(chunk)
-                        progress.update(len(chunk))
-                        downloaded_bytes += len(chunk)
-            # Return the number of bytes downloaded
-            return downloaded_bytes
+    def _download(self, url, path, session, file_size, product_info=None):
```
j08lue: You don't need the optional `product_info`. The function could also just return / append whatever a list of the `url`s that were passed in. In any case, the tracking of failure / success should happen outside the downloading function.

sentinelsat/sentinel.py
```diff
@@ -806,30 +885,43 @@ def _md5_compare(self, file_path, checksum, block_size=2 ** 13):
...
+                            progress.update(len(chunk))
+                            downloaded_bytes += len(chunk)
+                if product_info is not None and product_info['title'] in self.failed_list_title:
+                    self.failed_list_title.remove(product_info['title'])
+                    self.failed_list_id.remove(product_info['id'])
+
+                # Return the number of bytes downloaded
+                return downloaded_bytes
+        # meant for catching other types of exceptions such as 500 internal server side errors
+        except Exception as ex:
```
j08lue: Same as above. Never catch-all for a whole body of code. What if you introduced a typo in the code? You would not be able to tell whether the download failed because of SciHub / a cosmic ray flipping a bit or because your code broke...

sentinelsat/sentinel.py
ziya5635: To be honest, I intentionally did it since Fin-Hub has been suffering from two major issues, one was related to some internal server issues that make the code crashes, so I was forced to wrap the methods by an error catcher otherwise I would not be able to test their services at all. The other issue with their server is that some products are missing. They are found by queries but when an HTTP request is sent to pool the product, the product can not be found, which makes python crash.
These issues forced me to use some extra error catcher in different places.

sentinelsat/sentinel.py
ziya5635: As I explained above, there are two major issues with Fin-hub servers, so to test their server I had to catch the error raised by _download method. there was no need to do that if the server had not raised 500 internal errors! right now your library crashes using Fin-hub servers due to these issues. They have promised to solve this hopefully in this September.

.idea/misc.xml
ziya5635: I am happy that this functionality was added to this package. I do keep contributing. The reason for those extra error catchers is explained below. I had to do it that way to test Fin-hub servers.

sentinelsat/sentinel.py
ziya5635: yes sure, I remember that next time!

sentinelsat/sentinel.py
ziya5635: sure, I leave them to the users.

sentinelsat/sentinel.py
j08lue: I understand, thanks for the explanation.

Issue #276, opened 2019-04-09
Searching based on place names


valgur:
This is a feature idea I have had for some time but have not got around to implementing.

I propose we add a `--location` argument to the CLI that takes a natural-language place name as input. So instead of doing
```
sentinelsat -u <user> -p <password> -g <search_polygon.geojson> --sentinel 2 --cloud 30
```
you could do something like
```
sentinelsat -u <user> -p <password> --location Berlin --sentinel 2 --cloud 30
```

I would implement this by using one of the services supported via [geopy](https://github.com/geopy/geopy).

I would use intersection with the bounding box of the highest-ranked result as a search term in the SciHub API search query. Bounding box makes more sense for place names of larger regions, such as countries.
Maybe also support simple `lon,lat` or `lat,lon` directly, without a geocoding lookup. Which format to support is not a too obvious choice. Most geocoders use the `lat,lon` form but `lon,lat` would be more consistent with most GIS formats, including the GeoJSON files and WKT strings, so I'm leaning towards this one.

I would also add a small note of the location that was matched (e.g. its full place name) and its coordinates to the CLI output.
Adding this functionality to the sentinelsat's Python API as a separate function would make sense as well.

This feature would make sentinelsat much more straightforward to use for simple queries that only need products overlapping with some specific locations and would make it easier to provide examples in the readme and docs by not requiring a GeoJSON file to be created first.

Fernerkundung:
This proposal would definitely steer `sentinelsat` towards more user friendliness. I like it! :+1: 

The requirements added by `geopy` appear to to only be `geographiclib`, so we should be fine on the front of an easy install process on all plattforms and minimal technical debt.

Regarding point location coordinates I would use the `lon,lat` format to stay consistent with GIS formats and also to make parsing into a POINT WKT easier.

I would also add that exposing this to the Python API would be good, e.g. in the form of a function `wkt_from_location_name`.

j08lue:
Great enhancement idea.

Should we make the `geopy` and `geographiclib` dependencies optional, like we do with `geopandas`? Many API users might never need them. `geopy` is obviously much less heavy of a dependency than `geopandas`/`pandas`/`numpy`/`shapely`/whatnot, but still. Or would that get too complicated?

I also second the `wkt_from_location_name` idea.

dwlsalmeida:
I could lend a hand if you guys give some steering along the way.

Fernerkundung:
Please, do go ahead with this issue :+1: 

[...]

PR #309, opened 2019-08-18
Searching based on place names 

vlohiatest:
Referring #276:

Many geocoding services provided by Geopy require API tokens. I thought it would be useless to import a library for using only one of those services which doesn't require such tokens(like Nominatim). So i decided to directly use one of the supported services'(Nominatim) API directly in the
```naturalname_to_wkt```

The user will have to use it like this:
```
from sentinelsat import SentinelAPI, naturalname_to_wkt
api = SentinelAPI('user', 'passwd', 'https://scihub.copernicus.eu/dhus')
location, footprint=naturalname_to_wkt(naturalname="florida")
print("Geocoding result matched: ",location)
products = api.query(footprint,date=('20190810', '20190815'),Platformname='Sentinel-2')
```

### CODE REVIEW

File: sentinelsat/sentinel.py [hunk 1]
```diff
@@ -939,7 +939,40 @@ def check_bounds(geometry):
...
+    """Maps the naturalname to rectangular bounding extents using Nominatim API and converts to Well-Known-Text.
 
+    Parameters
+    ----------
+    naturalname : str
+	the query to geocode
+        
+    Returns
+    -------
+    polygon coordinates
```
Fernerkundung: The function does not return the polygon but a tuple of the placename and the coordinates. The preferable solution would be to mirror [`geojson_to_wkt`](https://github.com/sentinelsat/sentinelsat/blob/d9333fd92fc1fdf182bf0f55da34412586427ff3/sentinelsat/sentinel.py#L889) and just return the coordinates, so it can be a drop-in replacement.

i.e.
```
products = api.query(footprint=placename_to_wkt("florida"))
```

sentinelsat/sentinel.py
```diff
@@ -939,7 +939,40 @@ def check_bounds(geometry):
     wkt = re.sub(r'(?<!\d) ', '', wkt)
     return wkt
 
+def naturalname_to_wkt(naturalname):
```
Fernerkundung: I prefer `placename` instead of `naturalname`. This would also mirror the nominatim key `place` which will probably be the one the query hits most often.

sentinelsat/sentinel.py
vlohiatest: Yeah my bad! I will change it.
Another issue is: where the user should be intimated of the returned placename for which the query will execute. Isn't the user's approval required before executing the query as the placename may not be what he wants or simply be erroneous.

sentinelsat/sentinel.py
vlohiatest: ok!

sentinelsat/sentinel.py
Fernerkundung: For this feature I would prefer simplicity over explicitly knowing the place queried. The whole idea is to provide a simple search interface for people who do not want to deal with defining their area-of-interest with a polygon. So taking the first result in ambiguous cases should be fine.

If the returned bbox for a placename is ambiguous we could include this as a caveat in the docs or docstrings and point people to testing their search on nominatim first. If someone needs to be more explicit with what area they want to search providing a polygon is still the best option.

j08lue: I agree. Queries are cheap and people will hopefully check whether the results are as they expect before downloading all the data...

vlohiatest: ok i have made the changes.

sentinelsat/sentinel.py
```diff
@@ -26,7 +26,6 @@
 
 class SentinelAPI:
     """Class to connect to Copernicus Open Access Hub, search and download imagery.
-
```
Fernerkundung: Please do not remove the whitespace (here as well as in the rest of the code) that we used on purpose in the docstrings and throughout the rest of the code. I suspect this is some auto-formatting happening in your IDE. Please revert to the whitespace state of master.

sentinelsat/sentinel.py
```diff
@@ -939,24 +885,51 @@ def check_bounds(geometry):
...
+        #Get the First result's bounding box.
+        bbox= jsonlist['features'][0]['bbox']
+        #the bbox follows the pattern:
+        #[longitude lowerleft corner,latitude lowerleft corner,longitude upperright corner,latitude upperright corner]
+        footprint="POLYGON(({} {},{} {},{} {},{} {},{} {}))"
+        footprint=footprint.format(bbox[0],bbox[3],bbox[2],bbox[3], \
+                                   bbox[2],bbox[1],bbox[0],bbox[1], \
+                                   bbox[0],bbox[3])
+        return footprint
+    except:
```
Fernerkundung: Please try to catch a specific exception. With this statement you are blanket catching *all* exceptions and returning an IndexError, even though the source error might be something else.

sentinelsat/sentinel.py
```diff
@@ -939,24 +885,51 @@ def check_bounds(geometry):
...
+    polygon coordinates
+        string of comma separated coordinate tuples (lon, lat) to be used by SentinelAPI
+    """
+    rqst = requests.post('https://nominatim.openstreetmap.org/search', \
+                         params = { 'q':placename,'format':'geojson'})
+    #Check that the response from Openstreetmapserver has status code 2xx 
+    #and that the response is valid JSON.
+    _check_scihub_response(rqst)
+    try:
+        jsonlist=rqst.json()
```
Fernerkundung: Please use consistent spaces inbetween operators, i.e.: jsonlist = rqst.json() and further below bbox = jsonlist, footprint = "POLYGON... etc.

sentinelsat/sentinel.py
```diff
@@ -939,24 +885,51 @@ def check_bounds(geometry):
     wkt = re.sub(r'(?<!\d) ', '', wkt)
     return wkt
 
-
+def placename_to_wkt(placename):
+    """Maps the placename to rectangular bounding extents using Nominatim API and converts to Well-Known-Text.
+    Parameters
```
Fernerkundung: Include whitespace to be consistent with the other function definitions in the code.

sentinelsat/sentinel.py
j08lue: Yes. Which line can cause which exception? Wrap only that line in a `try ... except KeyError:` or `try ... except IndexError:` or what else it might be. Then you can me much more sure that the exception you catch is the one you intend to (and not something totally different that you misinterpret).

sentinelsat/sentinel.py
```diff
@@ -939,24 +885,51 @@ def check_bounds(geometry):
...
+        bbox= jsonlist['features'][0]['bbox']
+        #the bbox follows the pattern:
+        #[longitude lowerleft corner,latitude lowerleft corner,longitude upperright corner,latitude upperright corner]
+        footprint="POLYGON(({} {},{} {},{} {},{} {},{} {}))"
+        footprint=footprint.format(bbox[0],bbox[3],bbox[2],bbox[3], \
+                                   bbox[2],bbox[1],bbox[0],bbox[1], \
+                                   bbox[0],bbox[3])
+        return footprint
+    except:
+        raise IndexError("Empty search results for given Placename query!")
```
j08lue: This should perhaps be a `ValueError`: the value that the user provided (some unresolveable place name) was the source of failure.

vlohiatest:
Thank you for your reviews. I will share the updated code shortly.

PR #310, opened 2019-08-27
Searching based on place names

vlohiatest:
This is the new pull request in replacement of #309.
I have incorporated some of the suggested changes and here are some details which i think must be conveyed:

1.Exception handling in ```placename_to_wkt()```:
The Nominatim API seems to return a dictionary with empty array ```features``` (which has the bbox info) when no results are matched. For the exception to be captured i believe the error is an ```IndexError```. For the server-side invalid API responses and wrong geojson formats, i think ```_check_scihub_response()``` will effectively handle them without needing code duplication. The API also respects the lon,lat in ```bbox``` to be in -180,180 range and i could not find any client-side query constraints in their API documentation. I cannot think of any other exceptions that can arise so some suggestions can be helpful.

2.```test_placename_to_wkt()``` is included in ```test_geographic.py``` but i am unsure about my implementation since it compares the exact ```wkt``` returned. Here too suggestions are appreciated.

Thank You

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -1097,6 +1097,44 @@ def geojson_to_wkt(geojson_obj, feature_number=0, decimals=4):
     wkt = re.sub(r'(?<!\d) ', '', wkt)
     return wkt
 
+def placename_to_wkt(placename):
+    """Maps the placename to rectangular bounding extents using Nominatim API and converts to Well-Known-Text.
+   
+    Parameters
+    ----------
+    placename : str
+        the query to geocode
+    
+    Returns
+    -------
+    polygon coordinates
+        string of comma separated coordinate tuples (lon, lat) to be used by SentinelAPI
+    """
+
+    rqst = requests.post('https://nominatim.openstreetmap.org/search', \
+                         params = { 'q':placename,'format':'geojson'})
+    #Check that the response from Openstreetmapserver has status code 2xx 
+    #and that the response is valid JSON.
+    _check_scihub_response(rqst)
+    jsonlist=rqst.json()
+    try:
+        #Get the First result's bounding box.
+        bbox= jsonlist['features'][0]['bbox']
+    except IndexError:
+        #bbox is empty
+        raise Exception("No match found for entered Placename query!")
+    except:       
+        #unknown error
+        raise Exception("Unexpected error")
+    else:
+        #the bbox follows the pattern:
+        #[longitude lowerleft corner,latitude lowerleft corner,longitude upperright corner,latitude upperright corner]
+        footprint="POLYGON(({} {},{} {},{} {},{} {},{} {}))"
+        footprint=footprint.format(bbox[0],bbox[3],bbox[2],bbox[3], \
+                                   bbox[2],bbox[1],bbox[0],bbox[1], \
+                                   bbox[0],bbox[3])
+        return footprint
+
 
 def format_query_date(in_date):
     """
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -1116,10 +1116,10 @@ def placename_to_wkt(placename):
     #Check that the response from Openstreetmapserver has status code 2xx 
     #and that the response is valid JSON.
     _check_scihub_response(rqst)
-    jsonlist=rqst.json()
+    jsonlist = rqst.json()
     try:
         #Get the First result's bounding box.
-        bbox= jsonlist['features'][0]['bbox']
+        bbox = jsonlist['features'][0]['bbox']
     except IndexError:
         #bbox is empty
         raise Exception("No match found for entered Placename query!")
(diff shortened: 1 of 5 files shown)


sentinelsat/sentinel.py
```diff
@@ -1097,6 +1097,44 @@ def check_bounds(geometry):
...
+    except IndexError:
+        #bbox is empty
+        raise Exception("No match found for entered Placename query!")
+    except:       
+        #unknown error
+        raise Exception("Unexpected error")
+    else:
+        #the bbox follows the pattern:
+        #[longitude lowerleft corner,latitude lowerleft corner,longitude upperright corner,latitude upperright corner]
+        footprint = "POLYGON(({} {},{} {},{} {},{} {},{} {}))"
```
valgur: SciHub's Solr backend supports a much simpler `ENVELOPE(minX, maxX, maxY, minY)` WKT geometry, I'd suggest using it instead of `POLYGON((...))`. See [this test](https://github.com/sentinelsat/sentinelsat/blob/d9333fd92fc1fdf182bf0f55da34412586427ff3/tests/test_opensearch.py#L325) for an example.
Also, as a side note, `footprint:[minY,minX TO maxY,maxX]` works as an equivalent to `footprint:"Intersects(ENVELOPE(minX, maxX, maxY, minY))"`, curiously.

sentinelsat/sentinel.py
```diff
@@ -1097,6 +1097,44 @@ def check_bounds(geometry):
...
+   
+    Parameters
+    ----------
+    placename : str
+        the query to geocode
+    
+    Returns
+    -------
+    polygon coordinates
+        string of comma separated coordinate tuples (lon, lat) to be used by SentinelAPI
```
valgur: Need to update the docstring.

sentinelsat/sentinel.py
valgur: I would also unpack the `bbox` variable as `minx, miny, maxx, maxy = bbox` and use these to construct the WKT string for clarity.

vlohiatest: i will make the changes accordingly

vlohiatest:
I have updated the code. I don't know why some builds are failing and the code coverage is low. Should i be worried about this?

tests/test_geographic.py
```diff
@@ -6,7 +6,7 @@
 import geojson
 import pytest
 
-from sentinelsat import geojson_to_wkt, read_geojson, SentinelAPI
+from sentinelsat import geojson_to_wkt, read_geojson, SentinelAPI, placename_to_wkt
```
j08lue: Just for looks, put the `placename_to_wkt` after `geojson_to_wkt`.

tests/test_geographic.py
```diff
@@ -58,3 +58,14 @@ def test_footprints_s2(products, fixture_path):
...
     # to compare unordered lists (JSON objects) they need to be sorted or changed to sets
     assert set(footprints) == set(expected_footprints)
+    
+@pytest.mark.vcr
+@pytest.mark.fast
+def test_placename_to_wkt(place_kwargs):    
+    #tests wkt response
+    assert placename_to_wkt(place_kwargs[0])==place_kwargs[1]
+    #tests empty bbox exception in response to bad query
+    with pytest.raises(Exception) as excinfo:
```
j08lue: Rather specify which exception you expect, please.

sentinelsat/sentinel.py
```diff
@@ -1097,6 +1097,42 @@ def check_bounds(geometry):
...
+    #and that the response is valid JSON.
+    _check_scihub_response(rqst)
+    jsonlist = rqst.json()
+    try:
+        #Get the First result's bounding box.
+        bbox = jsonlist['features'][0]['bbox']
+    except IndexError:
+        #bbox is empty
+        raise Exception("No match found for entered Placename query!")
+    except:
```
valgur: Avoid overly broad exception catching. Re-raising with a generic error message does not add anything useful.

vlohiatest: I cannot think of any other scenario where any other exception can result since most of them are handled by _check_scihub_response(). Do you want to check for valid bbox coordinates too? I require help on this one.

j08lue:
> I have updated the code. I don't know why some builds are failing and the code coverage is low. Should i be worried about this?

Hmm. It is your test that is failing for [Python 3.4](https://travis-ci.com/sentinelsat/sentinelsat/jobs/229918827#L728) and [Python 3.5](https://travis-ci.com/sentinelsat/sentinelsat/jobs/229918828#L470) due to missing VCR cassettes.

It is because the tests send the request `https://nominatim.openstreetmap.org/search?format=geojson&q=florida` and the VCR cassette has `https://nominatim.openstreetmap.org/search?q=florida&format=geojson` https://github.com/sentinelsat/sentinelsat/pull/310/files#diff-ccfe41028e7e3e9a57c795631af1735fR16

Probably a problem of dictionaries not sorted before Python 3.6, which changes the order?

vlohiatest:
Can someone review the changes i have made?

Fernerkundung:
Closed as stale. If you want to continue this PR please comment and we can reopen it.

thomasyoung-audet:
I'd like to give this a shot. Could you reopen it?

Fernerkundung:
You should be able to checkout this PR to your own branch, continue development there and then create a new PR from that.

Issue #289, opened 2019-05-22
Required String parameter 'q' is not present


drnextgis:
I'm trying to run this simple command but it throws an error:

```
$ sentinelsat -u *** -p *** -g catalunya.geojson --sentinel 2 --cloud 30
Traceback (most recent call last):
  File "/home/denis/satellogic/env-shub/bin/sentinelsat", line 11, in <module>
    sys.exit(cli())
  File "/home/denis/satellogic/env-shub/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/denis/satellogic/env-shub/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/denis/satellogic/env-shub/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/denis/satellogic/env-shub/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/denis/satellogic/env-shub/lib/python3.6/site-packages/sentinelsat/scripts/cli.py", line 160, in cli
    order_by=order_by, limit=limit, **search_kwargs)
  File "/home/denis/satellogic/env-shub/lib/python3.6/site-packages/sentinelsat/sentinel.py", line 147, in query
    response, count = self._load_query(query, formatted_order_by, limit, offset)
  File "/home/denis/satellogic/env-shub/lib/python3.6/site-packages/sentinelsat/sentinel.py", line 272, in _load_query
    products, count = self._load_subquery(query, order_by, limit, offset)
  File "/home/denis/satellogic/env-shub/lib/python3.6/site-packages/sentinelsat/sentinel.py", line 304, in _load_subquery
    _check_scihub_response(response)
  File "/home/denis/satellogic/env-shub/lib/python3.6/site-packages/sentinelsat/sentinel.py", line 1027, in _check_scihub_response
    raise api_error
sentinelsat.sentinel.SentinelAPIError: HTTP status 400 Bad Request: 
# HTTP Status 400 - Required String parameter 'q' is not present

 **type** Status report

 **message** _Required String parameter 'q' is not present_

 **description** _The request sent by the client was syntactically incorrect._

* * *

### Apache Tomcat/8.0.36
```

drnextgis:
Sometimes for the same request I get the following error:

```
[...]

PR #318, opened 2019-10-03
Fix #290: Check query length in CLI and warn user

### LINKED ISSUE #290: Check query length in CLI and warn user
Warn users of the CLI that complex polygons might cause uncertain responses/crashes from SciHub. This entails minor additions to the CLI and the documentation.

* check if the query exceeds the proposed length limit with [api.check_query_length()](https://sentinelsat.readthedocs.io/en/stable/api.html#sentinelsat.SentinelAPI.check_query_length)
* provide a warning to the user that the query is complex and might cause a bad SciHub response
* extend documentation for issues caused by too complex polygons in queries (probably under [common issues](https://sentinelsat.readthedocs.io/en/stable/common_issues.html))

As proposed by @valgur in #289 after the report by @drnextgis.

yepremyana:
The following items are included in this PR

-  check if the query exceeds the 200 length limit with api.check_query_length()
- provide a logger warning to the user that the query is complex
- extend documentation for issues caused by too complex polygons in queries (under common issues)

fixes #290


j08lue:
Nice. But where did you get the 200 from? 😄 The return from the api function is a float between 0 and 1. So the condition should be `< 1`.

https://github.com/sentinelsat/sentinelsat/blob/80d01bdf28f5330e5515a6d77f4159dd0194198d/sentinelsat/sentinel.py#L813-L816

sentinelsat/scripts/cli.py
```diff
@@ -137,6 +137,8 @@ def cli(user, password, geometry, start, end, uuid, name, download, sentinel, pr
 
     if geometry is not None:
         search_kwargs['area'] = geojson_to_wkt(read_geojson(geometry))
+        if api.check_query_length(search_kwargs['area']) > 200:
```
j08lue: condition should be `check_query_length() >= 1`

sentinelsat/scripts/cli.py
```diff
@@ -137,6 +137,8 @@ def cli(user, password, geometry, start, end, uuid, name, download, sentinel, pr
...
+            logger.warning('Query request is too large, reduce query datapoints')
```
j08lue: Please change to `'Query is too complex. Consider using a simpler geometry.'`

sentinelsat/scripts/cli.py
yepremyana: I think there may be an error somewhere else then as well. I tested `api.check_query_length()` with the .geojson provided in #289 (which created the original error). And the output was 689.16937531742. Looking in the comments I noticed someone mentioned the number 200 max. I've never worked with .geojson before so would appreciate learning where I went wrong!

yepremyana: changed

sentinelsat/scripts/cli.py
yepremyana: changed

sentinelsat/scripts/cli.py
```diff
@@ -137,6 +137,8 @@ def cli(user, password, geometry, start, end, uuid, name, download, sentinel, pr
 
     if geometry is not None:
         search_kwargs['area'] = geojson_to_wkt(read_geojson(geometry))
+        if api.check_query_length(search_kwargs['area']) >= 1:
+            logger.warning('Query is too complex. Consider using a simpler geometry.')
```
j08lue: please raise a `ValueError` instead of warning.

yepremyana: will change to `ValueError`

yepremyana:
For the test, could we use the `run_cli` function with parameter `must_return_nonzero = True`?

j08lue:
> could we use the `run_cli` function with parameter `must_return_nonzero = True`

Sure, nice idea.

j08lue:
Don't know what this `must_return_nonzero` does, but it sure has some weird implications - a socket access error?! https://travis-ci.com/sentinelsat/sentinelsat/jobs/243519010

Maybe try the old way - get the return from `run_cli` and then check the error code yourself - if that is what works.

valgur:
> Don't know what this must_return_nonzero does, but it sure has some weird implications - a socket access error?! https://travis-ci.com/sentinelsat/sentinelsat/jobs/243519010

That's actually a useful and intentional error message, although it would have failed at a later step without it anyway. The added unit test is missing a corresponding `vcrpy` cassette.

j08lue:
I remember `vcrpy` errors to be more descriptive - like telling you that you ran without recording new cassettes and something was missing - or so... Well ¯\_(ツ)_/¯

tests/test_cli.py
```diff
@@ -136,6 +136,14 @@ def test_cli(run_cli, geojson_path):
 
     run_cli("--geometry", geojson_path, "-q", "producttype=GRD,polarisationmode=HH")
 
+@pytest.mark.fast
+def test_complex_geometry(run_cli, fixture_path):
+    with pytest.raises(ValueError):
+        run_cli(
+                '--geometry', fixture_path('map_complex_geometry.geojson'),
```
j08lue: We should probably not add a rather big GeoJSON file to this repo just for this test. Let us generate that dynamically in a fixture instead with some random coordinates. Ideally, this geometry never gets sent to the API endpoint anyways, since the validation you added kicks in before.

tests/test_cli.py
```diff
@@ -136,6 +136,14 @@ def test_cli(run_cli, geojson_path):
 
     run_cli("--geometry", geojson_path, "-q", "producttype=GRD,polarisationmode=HH")
```
j08lue: PEP8-format this, please.

PR #320, opened 2019-10-03
Fix filename for sentintel-5p

### LINKED ISSUE #270: Sentintel5 files are incorrectly downloaded as zip files when they are actually nc files
When I download Sentinel5 files they are downloaded as *.zip.  However they are not zip files, but are netcdf files and should have the extension *.nc.

lkiesow:
This patch fixes the issue that sentinel-5p files are downloaded as
*.zip files and not as *.nc files. Instead of hard-coding the filename,
sentinelsat will now make an HTTP head request to determine the name. In
case this request yields no result, the code will still fall back to the
old pattern.

This fixes #270

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -525,9 +525,17 @@ class SentinelAPI:
             If the MD5 checksum does not match the checksum on the server.
         """
         product_info = self.get_product_odata(id)
-        path = join(directory_path, product_info["title"] + ".zip")
-        product_info["path"] = path
-        product_info["downloaded_bytes"] = 0
+
+        # determine filename
+        req = self.session.head(product_info['url'], auth=self.session.auth)
+        filename = req.headers.get('Content-Disposition')
+        if filename:
+            filename = filename.split('=', 1)[1].strip('"')
+        else:
+            filename = product_info['title'] + '.zip'
+        path = join(directory_path, filename)
+        product_info['path'] = path
+        product_info['downloaded_bytes'] = 0
 
         self.logger.info("Downloading %s to %s", id, path)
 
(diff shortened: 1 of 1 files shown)


lkiesow:
Seems like the tests complain about the head requests. I will try to take a look at this over the next week if possible.

lkiesow:
I used a pragmatic solution to the problem and modified the previous patch so that sentinelsat uses the correct file extension for sentinel-5p files while not breaking half of the tests. Note that the previous solution (first commit) was the better one but would, unfortunately, need quite a bit of time for fixing the tests. If someone can spare that time, that would be great… but for now, this fixes the issue.

j08lue:
Thanks, @lkiesow, good work pointing out the two solutions. I am strongly in favor of reverting to your first commit and fixing the tests. Can you give me a heads-up to what needs to be done there? Just record new cassettes and tweak some fixtures or so? Or more involved changes?

lkiesow:
Should be the former. The way the first patch works is to make an HTTP HEAD request to determine the filename in advance. The tests do not expect this request to happen and hence fail.

craigmillernz:
Can this pull request be merged into master so that it's available on pip?

lkiesow:
Note that I'm not working on the discussed first solution since I have switched to the more specific [sentinel5dl](https://github.com/emissions-api/sentinel5dl) tool/lib. The current patch should still work like this though and would probably be helpful to some people. That said if you think it should be done properly, feel free to decline this pull request for now.

craigmillernz:
no problem i wasn't aware of that other tool, so I will investigate it.  The filename fix is fairly cosmetic as i can still read in the netcdf files even if they have a zip extension.

valgur:
This change was included in the merged #378. Thanks for the PR, nevertheless!

Issue #317, opened 2019-10-01
Add support for searching for multiple values of the same property in Python API


valgur:
Searching for multiple values of the same property simultaneously can be quite useful, e.g. for relative orbit numbers as discussed in https://github.com/sentinelsat/sentinelsat/issues/316#issuecomment-536984451.

We could add a better support for this by considering `api.query()` parameters of `set()` type as a set of allowed values. This avoids ambiguity with `tuple`/`list` with length of two that we interpret as a value range.

The set of values can then be translated to the raw query syntax using
```python
'{}:({})'.format(
    property_name, 
    ' OR '.join(map(str, sorted(values)))
)
```
So `api.query(relativeorbitnumber={50, 93})` produces `relativeorbitnumber:(50 OR 93)`, for example. The latter is equivalent to `relativeorbitnumber:50 OR relativeorbitnumber:93`, as described [here](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/query-dsl-query-string-query.html#_field_names).

OlgaCh:
@j08lue hello!
I'd like to work on this one as an hacktoberfest contribution.
Could you please assign me?

j08lue:
Thanks, @OlgaCh, I will. Please wait a day or two before you start hacking, though. We are just about to merge #296 that touches a lot of code and PRs branching off before that might cause some nasty merge conflicts.

OlgaCh:
Will do. Thank you for the headsup @j08lue

AmirM69:
> Searching for multiple values of the same property simultaneously can be quite useful, e.g. for relative orbit numbers as discussed in [#316 (comment)](https://github.com/sentinelsat/sentinelsat/issues/316#issuecomment-536984451).
> 
> We could add a better support for this by considering `api.query()` parameters of `set()` type as a set of allowed values. This avoids ambiguity with `tuple`/`list` with length of two that we interpret as a value range.
> 
> The set of values can then be translated to the raw query syntax using
> 
> ```python
> '{}:({})'.format(
>     property_name, 
>     ' OR '.join(map(str, sorted(values)))
> )
> ```
> 
[...]

PR #321, opened 2019-10-04
Allow to provide multiple values of property when build search query.

OlgaCh:
Closes https://github.com/sentinelsat/sentinelsat/issues/317

Enables an option to provide values for same parameter as a set.


OlgaCh:
@j08lue could you review this please?
Sorry for messy commit history. I've forget to run black before submitting PR and then have some issues dealing with it's warnings during the Travis build.

valgur:
Should also add a mention of this new functionality in documentation, but this can be addressed in another PR.

sentinelsat/sentinel.py
```diff
@@ -209,7 +209,9 @@ def format_query(area=None, date=None, raw=None, area_relation="Intersects", **k
             date_attrs = ["beginposition", "endposition", "date", "creationdate", "ingestiondate"]
             if attr.lower() in date_attrs:
                 # Automatically format date-type attributes
-                if isinstance(value, string_types) and " TO " in value:
+                if isinstance(value, set):
```
j08lue: @Fernerkundung and @valgur - only `set` should be coerced to `OR`? Not `tuple` and `list`, too?

j08lue: I guess it makes most sense for `set`, because it has unique elements and is unordered.

Fernerkundung: In the current state only sets are coerced as `OR`. Everything else gets parsed as `START TO END`.

Do we also want to support: `START1 TO END1 OR START2 TO END2`? That would require some changes to the logic. However, I can't see a compelling use case for this that can't be solved with 2 queries.

j08lue: no, fine this way then.

PR #327, opened 2019-10-26
Check query length in CLI and warn user 

sdf94:
Possible fix to #290 #318 
@Fernerkundung


j08lue:
Thanks. But don't we already have #318 for this?

sdf94:
I added the test, which did not seem like it was added to the code. Correct me if I am wrong? :)

Issue #311, opened 2019-08-30
Could not download OLCI from https://coda.eumetsat.int


paalge:
Hi

I'm new to using sentinelsat.
I'm currently running the master branch (v0.13-37-g533e243) of sentinelsat (due to this issue #294)  and trying to download OLCI data from eumetsat.
The command is this:

`sentinelsat --url https://coda.eumetsat.int  -u uname -p 'password' -g some.json --sentinel 3 -s 20190515 -e 20190518 --instrument OLCI  -d`

The output I get is this:

Found 25 products
Will download 25 products using 2 workers
```
Traceback (most recent call last):
  File "/home/pal/.local/bin/sentinelsat", line 11, in <module>
    load_entry_point('sentinelsat==0.13', 'console_scripts', 'sentinelsat')()
  File "/home/user/.local/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/user/.local/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/user/.local/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/user/.local/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/user/.local/lib/python3.6/site-packages/sentinelsat/scripts/cli.py", line 168, in cli
    product_infos, triggered, failed_downloads = api.download_all(products, path)
  File "/home/user/.local/lib/python3.6/site-packages/sentinelsat/sentinel.py", line 690, in download_all
    raise next(iter(x.exception() for x in dl_tasks if x.exception()))
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/user/.local/lib/python3.6/site-packages/sentinelsat/sentinel.py", line 761, in _download_online_retry
    if self.is_online(product_info['id']):
  File "/home/user/.local/lib/python3.6/site-packages/sentinelsat/sentinel.py", line 457, in is_online
    raise SentinelAPIError('Could not verify whether product {} is online'.format(id), r)
sentinelsat.sentinel.SentinelAPIError: HTTP status 404 Not Found: Could not verify whether product 4c6186dd-f89e-4164-86bb-b597a478821a is online
```

The json file is a 5 point polygon.

[...]

PR #342, opened 2020-01-31
added DHuS version property

gbaier:
Added a property that returns the DHuS version string. First step to address #311.

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -79,9 +79,24 @@ class SentinelAPI:
         self.session.headers["User-Agent"] = self.user_agent
         self.show_progressbars = show_progressbars
         self.timeout = timeout
+        self._dhus_version = None
         # For unit tests
         self._last_query = None
         self._last_response = None
+
+    @staticmethod
+    def _api2dhus_url(api_url):
+        return api_url.replace("apihub", "dhus")
+
+    @property
+    def dhus_version(self):
+        if self._dhus_version is None:
+            dhus_url = self._api2dhus_url(self.api_url)
+            resp = self.session.get(dhus_url + "api/stub/version",
+                                    auth=self.session.auth,
+                                    timeout=self.timeout)
+            self._dhus_version = resp.json()['value']
+        return self._dhus_version
 
     def query(
         self,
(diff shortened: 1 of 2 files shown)


valgur:
I checked their source code to see if it is indeed only available via the DHuS endpoint and not apihub. Appears to be so: https://github.com/SentinelDataHub/dhus-ajs/tree/master/src/main/java/fr/gael/dhus/server/http/webapp/stub/controller

sentinelsat/sentinel.py
```diff
@@ -79,10 +79,25 @@ def __init__(
...
         self._last_response = None
 
+    @staticmethod
+    def _api2dhus_url(api_url):
+        return api_url.replace("apihub", "dhus")
+
+    @property
+    def dhus_version(self):
+        if self._dhus_version is None:
+            dhus_url = self._api2dhus_url(self.api_url)
```
Fernerkundung: Since there is only 1 "apihub" and multiple "dhus" I think it would be nicer to have this in a `try...except` block, i.e. try to get the version first and then call ` _api2dhus_url`. 

I know in reality it does not make much of a difference but from a code clarity standpoint I think it increases the readability if we treat the exception to the rule as an actual `exception`.

Fernerkundung:
Thank you for the PR and testing this so thoroughly :+1: 

The `DHuS` version stub seems to be universal for all the national hubs.
- http://collaborative.mt.asi.it/api/stub/version
- https://colhub.met.no/api/stub/version
- https://ipsentinel.ipma.pt/dhus/api/stub/version
- https://data.sentinel.zamg.ac.at/api/stub/version
- https://finhub.nsdc.fmi.fi/api/stub/version
- https://sentinels.space.noa.gr/dhus/api/stub/version

sentinelsat/sentinel.py
Fernerkundung: Ignore the Python 3.4 build error for now.It is a dependency issue that we need to resolve.

gbaier: > Since there is only 1 "apihub" and multiple "dhus" I think it would be nicer to have this in a `try...except` block, i.e. try to get the version first and then call ` _api2dhus_url`.

I wrapped the call in a `try...except` block. Although, in my opinion the code is now more verbose and less readable. But it is also not a big deal. I'm fine with both versions.

PR #367, opened 2020-06-11
Add CLI flag for DHuS version

### LINKED ISSUE #350: Display DHuS version in CLI
Beginner friendly feature:

- #342 introduced a function to get the DHuS server-side version
- add a command-line entry with `click` to surface that version to CLI users
- name to be discussed, e.g. `sentinelsat --dhus-version`

Discussion:
- Is this too specialized to be added to the CLI and just clutters the arguments?
- Could this be aggregated to a `--info` argument that shows an overview of information of the DHuS server, capabilities, LTA, etc.?

thomasyoung-audet:
Closes #350 

- [x] Flag
- [x] Tests

### DIFF
--- a/sentinelsat/scripts/cli.py
+++ b/sentinelsat/scripts/cli.py
@@ -93,10 +93,7 @@ class CommaSeparatedString(click.ParamType):
     help="Limit search to a specific instrument on a Sentinel satellite.",
 )
 @click.option(
-    "--producttype",
-    type=str,
-    default=None,
-    help="Limit search to a Sentinel product type."
+    "--producttype", type=str, default=None, help="Limit search to a Sentinel product type."
 )
 @click.option(
     "-c",
@@ -136,13 +133,7 @@ class CommaSeparatedString(click.ParamType):
     and metadata of the returned products.
     """,
 )
-
-@click.option(
-    "--dhusversion",
-    is_flag=True,
-    is_eager=True,
-    help="displays the DHuS version used"
-)
+@click.option("--dhusversion", is_flag=True, is_eager=True, help="displays the DHuS version used")
 @click.version_option(version=sentinelsat_version, prog_name="sentinelsat")
 def cli(
     user,
@@ -170,6 +161,7 @@ def cli(
     Beyond your Copernicus Open Access Hub user and password, you must pass a geojson file
     containing the geometry of the area you want to search for or the UUIDs of the products. If you
     don't specify the start and end dates, it will search in the last 24 hours.
+    Can also provide the Copernicus Open Access Hub (DHuS) version.
     """
 
     _set_logger_handler()
--- a/sentinelsat/scripts/cli.py
+++ b/sentinelsat/scripts/cli.py
@@ -30,16 +30,6 @@ class CommaSeparatedString(click.ParamType):
             return value
 
 
-def print_dhus_version(ctx, value):
-    if not value or ctx.resilient_parsing:
-        return
-    API = SentinelAPI("tyoungaudet", "Fu85N8TcVCWi")
-    click.echo(API.dhus_version)
-    click.echo(API.dhus_version)
-    click.echo(str(type(API.dhus_version)))
-    ctx.exit()
-
-
 @click.command(context_settings=dict(help_option_names=["-h", "--help"]))
 @click.option(
     "--user",
@@ -146,17 +136,14 @@ def print_dhus_version(ctx, value):
     and metadata of the returned products.
     """,
 )
+
 @click.option(
-    "--dhus-version",
+    "--dhusversion",
     is_flag=True,
-    callback=print_dhus_version,
-    expose_value=False,
     is_eager=True,
     help="displays the DHuS version used"
 )
 @click.version_option(version=sentinelsat_version, prog_name="sentinelsat")
-
-
 def cli(
     user,
     password,
@@ -176,7 +163,7 @@ def cli(
     url,
     order_by,
     limit,
-
+    dhusversion,
 ):
     """Search for Sentinel products and, optionally, download all the results
     and/or create a geojson file with the search result footprints.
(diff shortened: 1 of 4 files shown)


sentinelsat/scripts/cli.py
```diff
@@ -133,6 +133,7 @@ def convert(self, value, param, ctx):
     and metadata of the returned products.
     """,
 )
+@click.option("--dhusversion", is_flag=True, is_eager=True, help="displays the DHuS version used")
```
Fernerkundung, line 136: I would prefer a more generic `info` option. This would allow extension of other DHuS information we want to surface later.

sentinelsat/scripts/cli.py
```diff
@@ -153,12 +154,14 @@ def cli(
...
     order_by,
     limit,
+    dhusversion,
 ):
     """Search for Sentinel products and, optionally, download all the results
     and/or create a geojson file with the search result footprints.
     Beyond your Copernicus Open Access Hub user and password, you must pass a geojson file
     containing the geometry of the area you want to search for or the UUIDs of the products. If you
     don't specify the start and end dates, it will search in the last 24 hours.
+    Can also provide the Copernicus Open Access Hub (DHuS) version.
```
Fernerkundung, line 164: Can be removed from the docstring - the help text of the click decorator is sufficient info.

sentinelsat/sentinel.py
```diff
@@ -91,15 +91,18 @@ def _api2dhus_url(api_url):
...
         try:
             resp = self.session.get(
-                self.api_url + "api/stub/version", auth=self.session.auth, timeout=self.timeout
+                self._api2dhus_url(self.api_url) + "api/stub/version",
+                auth=self.session.auth,
+                timeout=self.timeout,
             )
             resp.raise_for_status()
         except requests.exceptions.HTTPError as err:
+            # is this part of the code useful?
```
Fernerkundung, line 100: Please remove - a question is better suited as a comment on your PR instead of in the code.

sentinelsat/sentinel.py
```diff
@@ -91,15 +91,18 @@ def _api2dhus_url(api_url):
     def _req_dhus_stub(self):
         try:
             resp = self.session.get(
-                self.api_url + "api/stub/version", auth=self.session.auth, timeout=self.timeout
+                self._api2dhus_url(self.api_url) + "api/stub/version",
+                auth=self.session.auth,
+                timeout=self.timeout,
```
Fernerkundung, line 96: What is the motivation for changing this?

sentinelsat/sentinel.py
```diff
@@ -91,15 +91,18 @@ def _api2dhus_url(api_url):
...
             )
             resp.raise_for_status()
         except requests.exceptions.HTTPError as err:
+            # is this part of the code useful?
             self.logger.error("HTTPError: ", err)
             self.logger.error("Are you trying to get the DHuS version of APIHub?")
             self.logger.error("Trying again after conversion to DHuS URL")
             resp = self.session.get(
-                self._dhus2api_url(self.api_url) + "api/stub/version",
+                self._api2dhus_url(self.api_url) + "api/stub/version",
```
Fernerkundung, line 105: What is the motivation for changing the order of the name?

sentinelsat/sentinel.py
thomasyoung-audet: Removing.

sentinelsat/scripts/cli.py
thomasyoung-audet: Removing.

sentinelsat/scripts/cli.py
thomasyoung-audet: Done.

sentinelsat/sentinel.py
thomasyoung-audet: There was an error in the code. The function that we try to call located on line 88 is called `_api2dhus_url`, not `_dhus2api_url`.

sentinelsat/sentinel.py
thomasyoung-audet: This part of the code used to try to get a response from scihub.copernicus.eu/apihub/api/stub/version. This always returned a 404. Only after handling that 404 would the code try to get a response from scihub.copernicus.eu/dhus/api/stub/version. 
I don't know what the motivation for that initial attempt was, but It seems useless to me. That's why now it tries the dhus url from the start. 
Also there was a problem with the handling of the 404 error, something about logger not converting everything to strings that I couldn't quite figure out. So actually the error handler ran into an error and it was all a big mess. 
I was going to look into it more but I wanted to get the CLI working first.

thomasyoung-audet:
Hey could someone review this?

sentinelsat/sentinel.py
j08lue: The rationale for trying the original URL first would be that users can provide e.g. an API URL to a national DHuS endpoint or so. If that URL for some reason includes the string `apihub` anywhere, it will get replaced, which might lead to a URL that does not even exist. So the original behavior, trying the original URL first, is better. Could you please revert this?

I think the conflicting line about logging is this one:
```python
self.logger.error("HTTPError: ", err)
```

When you change that to
```python
self.logger.error("HTTPError: %s", err)
```
it should work.

valgur:
More of a nitpick: please follow these common conventions for git commit messages:
- https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53
- https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project

thomasyoung-audet:
> 
> 
> More of a nitpick: please follow these common conventions for git commit messages:
> 
>     * https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53
> 
>     * https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project

My apologies its my first time contributing. Will keep in mind for any future commits.

j08lue:
Great work! From my desk, this looks ready for merging. I wonder why the automatic black formatting is not fixing the code such that also the last CI job fails. Perhaps you should merge `master` into your branch here? Or we could just leave it and the CI should fix it after merging. I'll leave this to one of the other maintainers to decide, @valgur or @Fernerkundung.

Fernerkundung:
LGTM

Thanks for providing this PR and answering all comments @thomasyoung-audet 👍 

I'll populate the changelog and adapt the docs today and will merge it this afternoon.

PR #369, opened 2020-06-12
Reorder changelog items a bit

valgur:
Thanks for publishing a new release, @Fernerkundung!
Just a minor edit to the changelog. I think the changelog should be oriented towards end-users first, developers second. So, I suggest moving the development and testing change notes to the separate dedicated section.


Fernerkundung:
Agreed :+1: 

Thanks for bringing attention to this.

Issue #355, opened 2020-03-13
CLI should support geometry input as string


Fernerkundung:
Currently the `sentinelsat` CLI onyl supports input of search geometries as GeoJSON files.

**Proposal**
The CLI should accept strings of geometric objects, in addition to file paths. If the CLI supports direct input of WKT and/or GeoJSON strings, then you can chain it with any tool you like. (e.g. by chaining Fionas CLI fio (https://fiona.readthedocs.io/en/latest/cli.html) `fio info`, which returns a GeoJSON.) This would make the use of the CLI much more flexible and allow interconnection with different tools or the use in runtimes, where no "geometry" files can be created.

This was enhancement proposal was sparked by discussions in #343 .

**Design Considerations and Challenges**
Ideally the CLI accepts GeoJSON filepaths, GeoJSON strings or WKT Strings under the `-g, --geometry` flag.

The challenge is detecting the input format. In the API we have everything in place to parse GeoJSON filepaths or strings. WKT could be passed through without conversion.

j08lue:
> The challenge is detecting the input format.

If we want to handle this with a single flag (`--geometry`), we could have the heuristics like this
1. check whether the value is an existing path
2. check whether parsing it as JSON succeeds
3. treat it as WKT

Fernerkundung:
The Docs need an example of how to pipe WKT or GeoJSON strings from arbitrary sources into `sentinelsat`, e.g. by using [Fionas `fio info`](https://fiona.readthedocs.io/en/latest/cli.html#info).

thomasyoung-audet:
It looks to me like `fio info` doesn't exactly output a GeoJSON, but just a JSON object with information about the dataset. `fio dump` is the command that outputs a GeoJSON. `fio` does have a paramter `bounds` thought that we could maybe use? That's all we really need right?

thomasyoung-audet:
Anyone want to take a look at the pull request?

valgur:
Implemented in #381.

PR #372, opened 2020-06-22
Searching based on place name

### LINKED ISSUE #276: Searching based on place names

thomasyoung-audet:
Adds the functionality of doing a quick search and getting data based on a common place name.

Closes #276


tests/test_cli.py
```diff
@@ -528,3 +528,15 @@ def test_info_cli(run_cli, tmpdir):
...
     )
     tmpdir.remove()
+
+
+@pytest.mark.vcr
+@pytest.mark.scihub
+def test_location_cli(run_cli, tmpdir):
+    result = run_cli("--location", "Metz", "-l", "5")
+    assert (
+        "Found 23106 products\nThe location we are querying is: 'Metz, Moselle, Grand Est, France métropolitaine, "
```
j08lue, line 538: You can be a bit less specific on the test success criterion. E.g. just verify that "Found" is in the output, not the full message. Seems like your test is currently failing on some details in the text.

thomasyoung-audet:
Is it possible to rerun travis builds somehow because they are annoyingly inconsistent. Looking at the logs it doesn't seem to be making much sense.

valgur:
> Is it possible to rerun travis builds somehow because they are annoyingly inconsistent. Looking at the logs it doesn't seem to be making much sense.

Our VCR.py config needs some adjustment. It currently fails if GET query param order in the new query and the recorded cassette are different. Will have to make the matching a bit more fine-grained.

sentinelsat/sentinel.py
```diff
@@ -1464,7 +1464,9 @@ def placename_to_wkt(placename):
     jsonlist = rqst.json()
     try:
         # Get the First result's bounding box and description.
-        placeinfo = jsonlist["features"][0]["properties"]["display_name"]
+        placeinfo = jsonlist["features"][0]["properties"]["display_name"] + ". Coordinates: " + \
```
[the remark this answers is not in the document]
thomasyoung-audet: I'll get rid of it if you think its overkill, I was just going off of the requests made in https://github.com/sentinelsat/sentinelsat/issues/276.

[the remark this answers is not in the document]
j08lue: Let's ask @valgur or @Fernerkundung on a second opinion. Sometimes some extra lines of code are a good investment into good UX...

sentinelsat/sentinel.py
```diff
@@ -1438,3 +1438,40 @@ def _parse_odata_response(product):
...
+
+    Parameters
+    ----------
+    placename : str
+        the query to geocode
+
+    Returns
+    -------
+    full name and coordinates of the queried placename
+        list in the form [string placeinfo, wkt string in form 'ENVELOPE(minX, maxX, maxY, minY)']
```
valgur, line 1455: The order of return values in docs does not match the actual order.

sentinelsat/sentinel.py
```diff
@@ -1438,3 +1438,40 @@ def _parse_odata_response(product):
...
+    """
+
+    rqst = requests.post(
+        "https://nominatim.openstreetmap.org/search", params={"q": placename, "format": "geojson"}
+    )
+    # Check that the response from Openstreetmapserver has status code 2xx
+    # and that the response is valid JSON.
+    _check_scihub_response(rqst)
```
valgur, line 1463: Should simply use `rqst.raise_for_status()` here since the extra error handling included in `_check_scihub_response()` is not relevant for Nominatim.

sentinelsat/sentinel.py
```diff
@@ -1438,3 +1438,40 @@ def _parse_odata_response(product):
...
+        bbox = jsonlist["features"][0]["bbox"]
+
+    except IndexError:
+        # bbox or placeInfo is empty
+        raise ValueError('Unable to find a location for "{}"'.format(placename))
+    else:
+        # the bbox follows the pattern:[minX, minY, maxX ,maxY]
+        minX, minY, maxX, maxY = bbox
+        footprint = "ENVELOPE({}, {}, {}, {})".format(minX, maxX, maxY, minY)
+        return [footprint, placeinfo]
```
valgur, line 1477: Don't need the brackets here. Plain `footprint, placeinfo` is interpreted as a tuple.

valgur, line 1477: I would replace the try-catch with a simple check:
```python
    rqst.raise_for_status()
    jsonlist = rqst.json()
    if len(jsonlist["features"]) == 0:
        raise ValueError('Unable to find a matching location for "{}"'.format(placename))
    # Get the First result's bounding box and description.
    feature = jsonlist["features"][0]
    placeinfo = feature["properties"]["display_name"]
    minX, minY, maxX, maxY = feature["bbox"]
    footprint = "ENVELOPE({}, {}, {}, {})".format(minX, maxX, maxY, minY)
    return footprint, placeinfo
```

tests/conftest.py
```diff
@@ -188,3 +188,8 @@ def large_query():
         area="POLYGON((0 0,0 10,10 10,10 0,0 0))",
         date=(datetime(2015, 12, 1), datetime(2015, 12, 31)),
     )
+
+
+@pytest.fixture(scope="session")
+def place_kwargs():
+    return ["florida", "ENVELOPE(-87.634896, -79.974306, 31.000968, 24.396308)"]
```
valgur, line 195: I would just this value instead of using a fixture, since it is used only once and adds an unnecessary indirection.

tests/fixtures/vcr_cassettes/test_location_cli.yaml
```diff
@@ -0,0 +1,362 @@
...
+      date:
+      - Tue, 23 Jun 2020 16:32:27 GMT
+      pragma:
+      - no-cache
+      server:
+      - Apache-Coyote/1.1
+      set-cookie:
+      - dhusAuth=ea1a3b69907ce71cfb5090a665dfccf1; Path=/; HttpOnly
+      - dhusIntegrity=0b74759a8fe713f5e7b28bc2ddd71975269dc3e6; Path=/; HttpOnly
+      - JSESSIONID=5F6B9F9DF5B8C796F6AEFA7BC358EA9A; Path=/; HttpOnly
```
valgur, line 308: VCR.py changed their header naming to lower-case, breaking our scrubbing of cookies in cassettes. I fixed it in #375, but it has not been merged yet.
Please re-recorded the cassette with these fixes when they get merged. Even better if you can rebase your git history to purge this info from the repo.

tests/test_cli.py
```diff
@@ -528,3 +528,11 @@ def test_info_cli(run_cli, tmpdir):
         "DHuS version: 2.4.1\n" in result.output
     )
     tmpdir.remove()
+
+
+@pytest.mark.vcr
+@pytest.mark.scihub
+def test_location_cli(run_cli, tmpdir):
```
valgur, line 535: No need for including the `tmpdir` fixture here.

tests/test_geographic.py
```diff
@@ -58,3 +58,19 @@ def test_footprints_s2(products, fixture_path):
         expected_footprints = geojson.loads(geojson_file.read())
     # to compare unordered lists (JSON objects) they need to be sorted or changed to sets
     assert set(footprints) == set(expected_footprints)
+
+
+@pytest.mark.vcr
+@pytest.mark.fast
```
valgur, line 64: `@pytest.mark.scihub` marker would be a better fit here and in the next test. `fast` is intended for tests that don't try to access the internet.

sentinelsat/scripts/cli.py
```diff
@@ -231,6 +239,11 @@ def cli(
         with open(os.path.join(path, "search_footprints.geojson"), "w") as outfile:
             outfile.write(gj.dumps(footprints_geojson))
 
+    if location is not None:
+        query_answers = placename_to_wkt(location)
```
valgur, line 243: `wkt, place_name = placename_to_wkt(location)` would be a bit clearer.

sentinelsat/scripts/cli.py
```diff
@@ -231,6 +239,11 @@ def cli(
...
+        click.echo("The location we are querying is: '{}'".format(query_answers[1]))
```
valgur, line 244: I would also print the WKT envelope string in addition to the place name to make it clearer what the area being queried actually is and how big it is.

sentinelsat/sentinel.py
thomasyoung-audet: woops haha

sentinelsat/sentinel.py
```diff
@@ -1452,26 +1452,21 @@ def placename_to_wkt(placename):
...
-        minX, minY, maxX, maxY = bbox
-        footprint = "ENVELOPE({}, {}, {}, {})".format(minX, maxX, maxY, minY)
-        return [footprint, placeinfo]
+    if len(jsonlist["features"]) == 0:
+        raise ValueError('Unable to find a matching location for "{}"'.format(placename))
+    # Get the First result's bounding box and description.
+    feature = jsonlist["features"][0]
+    minX, minY, maxX, maxY = feature["bbox"]
+    footprint = "ENVELOPE({}, {}, {}, {})".format(minX, maxX, maxY, minY)
+    placeinfo = feature["properties"]["display_name"] + ", " + footprint
```
valgur, line 1471: Do the concatenation of `display_name` and `footprint` in `cli.py` instead of here please.

valgur:
Please merge `master` into your branch. This should get rid of the unit test failures in CI.

tests/fixtures/vcr_cassettes/test_location_cli.yaml
thomasyoung-audet: Rerecording the test and ran into a strange error with vcr itself. It seems to run into a problem while creating the yaml file, some dictionary doesn't have a matching key. I've attached a screenshot of the output. Have any of you had this problem before?
Screenshot from 2020-06-29 10-00-40

thomasyoung-audet: I've noticed its also the error that caused the CI to fail on the merge commit

valgur: That's a bug, indeed. Added a fix in #379.

thomasyoung-audet: done

thomasyoung-audet:
The pytests all pass on my end now. Travis fails because of some ascii problem.

Fernerkundung:
LGTM!

One final thing: Can you please populate the Changelog.rst with your additions @thomasyoung-audet and add yourself to the Contributors.rst document.

I think 1 bullet point in the added category would be fine with a reference to the PR number, i.e. "added searching by placenames (#372 @thomasyoung-audet).

valgur:
We have not decided to drop Python 2 support yet, so to fix the Python 2's Unicode handling error in the CI, please add a `u` prefix to the string in:
```python
click.echo(u"The location we are querying is: '{}', '{}'".format(place_name, wkt))
```

sentinelsat/scripts/cli.py
```diff
@@ -231,6 +244,11 @@ def cli(
         with open(os.path.join(path, "search_footprints.geojson"), "w") as outfile:
             outfile.write(gj.dumps(footprints_geojson))
 
+    if location is not None:
+        wkt, place_name = placename_to_wkt(location)
+        click.echo("The location we are querying is: '{}', '{}'".format(place_name, wkt))
+        search_kwargs["area"] = wkt
+
```
valgur, line 251: Oh, one miiiinor issue still... this piece of code has no effect since it is run *after* the query.
Please move it before
```python
    if geometry is not None:
        search_kwargs["area"] = geojson_to_wkt(read_geojson(geometry))
```
so the `--geometry` option takes precedence if both it and `--location` are set.

thomasyoung-audet: for geometry to take precedence shouldn't it be before location?

valgur: With it before `--geometry`, if both are set then the value from `--geometry` is the final one applied to `search_kwargs["area"]`, so its the one that gets used and takes precedence.

thomasyoung-audet: Ah I see ok

j08lue:
Wow, what a ride! Thanks for your stamina, @thomasyoung-audet, and the great support @valgur. This might become a very popular feature. Merging now.

Fernerkundung:
side note: With such a great new feature - should we package a new (minor) release soon?

thomasyoung-audet:
How often do you package releases? I am going to work on https://github.com/sentinelsat/sentinelsat/issues/355 this week, would you want that in the release?

j08lue:
> How often do you package releases?

To be honest, we just release when ever it feels right, usually a little late. 😄 But very good call to prevent that this time, @Fernerkundung! 

> would you want that in the release?

Yes, please let's include the `--geometry` improvement, since it is another improvement of the geo query UX.

PR #375, opened 2020-06-26
Fix unit test regressions

valgur:
* Fixes geopandas deprecation warnings for CRS format definition and missing geometry input.
* Fixes pytest warnings about missing marker definitions.
* Re-enable Unicode handling unit test as the [blocking VCR.py issue has been fixed](https://github.com/kevin1024/vcrpy/pull/386).
* Fix VCR.py header handling regression, which broke scrubbing of cookies (affects #372) and handling of downloaded files in cassettes.


Fernerkundung:
@valgur would you mind adding these changes to the changelog as well? Otherwise I fear we lose track of all the changes done inbetween releases.

PR #378, opened 2020-06-29
Better filename handling, exception info for failed download_all() products

### LINKED ISSUE #270: Sentintel5 files are incorrectly downloaded as zip files when they are actually nc files

valgur:
Continues from #320.
Tries to get the correct file name from the Content-Disposition header if possible, but guesses it if the header is unavailable.
I'm not 100% sure whether trying to use the header is a good idea, since it is not provided for archived products, which is a rather common case, and it adds an additional HEAD request. For archived products, the file name is needed to check if the file already exists on disk and can be ignored. Fixes #270 

Since this filename handling does an extra HEAD request for each product, the cassettes for all download tests needed updating, which I did.

The second main addition in this PR is the "exception" field to the `failed_products` dict returned by in `.download_all()`. Mostly to make debugging unit tests easier, but it seems like something potentially useful for end-users also. The only issue it might cause for users is when someone has decided to save the `failed_products` directly as JSON for some reason, but I don't see that as too likely and would be easy to fix.

Also tidied the download logic a bit:
* avoid unnecessary .is_online() query
* add max_attempts > 0 check
* reduce nesting
* merge error handling


Fernerkundung:
This looks really good @valgur :+1: 

I would like to provide a review on this but am limited on time until next week. If this is blocking too much down the line please feel free to go ahead without me.

valgur:
Making the file name query a boolean option might be worthwhile, but adding an option for the file extension seems like an overkill to me. If there's an actual need for this, then I would suggest the user to either overload or monkey patch the `._get_filename()` function (and to open a PR).

Also, TODO:
* [ ] a bunch of changelog entries.

tests/test_download.py
```diff
@@ -183,27 +184,28 @@ def test_download_all_one_fail(api, tmpdir, smallest_online_products):
...
+        assert sorted(failed_downloads) == [ids[0]], exceptions
+        assert type(list(exceptions.values())[0]) == InvalidChecksumError, exceptions
+        assert triggered == {}
+        assert sorted(list(product_infos) + list(failed_downloads)) == sorted(ids)
         assert id in failed_downloads
 
     tmpdir.remove()
 
 
+@pytest.mark.xfail(reason="The threading in this test seems to break VCR.py somehow")
```
j08lue, line 197: Hmm. We should fix this somehow, but that is a different issue to raise. Perhaps we can avoid Threading if `n_concurrent_dl` is 1.

Fernerkundung:
Any chance we can get this merged for a minor v0.14.1 release?

tests/test_download.py
j08lue: Let's revisit this if it pops up elsewhere later.

j08lue:
@valgur So much good work in this one that we should merge. To wrap this one up, I suggest we just add the CANGELOG entries and leave the additional unit test for another PR.

Suggested CHANGELOG additions:
* Get file name extension from header or internal logic (in particular for S5 products #270)
* Added an "exception" field to the `failed_products` dict returned by in `.download_all()`
* Tidied up the download logic

avalentino (approved):

sentinelsat/sentinel.py
```diff
@@ -572,11 +572,11 @@ def download(self, id, directory_path=".", checksum=True, **kwargs):
...
 
         .. versionchanged:: 0.15
-           Added ``**kwargs`` parameter to allow easier specialization of the :class:`SentinelAPI` cass.
+           Added ``**kwargs`` parameter to allow easier specialization of the :class:`SentinelAPI` class.
         """
         product_info = self.get_product_odata(id)
-        ext = kwargs.get("file_ext", ".zip")
-        path = join(directory_path, product_info["title"] + ext)
+        filename = self._get_filename(product_info)
+        path = join(directory_path, filename)
```
avalentino, line 579: OK, this will require a change in the PR #441 "New specialized client for the GNSS end-point on scihub"

avalentino (approved):

PR #381, opened 2020-07-07
Add CLI support for geometry input as string

thomasyoung-audet:
Adds CLI support for geometry input as string


sentinelsat/scripts/cli.py
```diff
@@ -68,9 +78,7 @@ def convert(self, value, param, ctx):
     show_default=True,
     help="End date of the query in the format YYYYMMDD.",
 )
-@click.option(
-    "--geometry", "-g", type=click.Path(exists=True), help="Search area geometry as GeoJSON file."
-)
+@click.option("--geometry", "-g", type=str, help="Search area geometry as GeoJSON file.")
```
Fernerkundung, line 81: The help string needs to be extended to reflect the change that now also GeoJSON strings and WKT strings are supported.

sentinelsat/sentinel.py
```diff
@@ -1457,3 +1457,11 @@ def placename_to_wkt(placename):
     footprint = "ENVELOPE({}, {}, {}, {})".format(minX, maxX, maxY, minY)
     placeinfo = feature["properties"]["display_name"]
     return footprint, placeinfo
+
+
+def is_wkt(possible_wkt):
+    SET_OF_ACCEPTABLE_SHAPES = "POLYGON((", "MULTIPOLYGON(((", "ENVELOPE("
```
Fernerkundung, line 1463: According to the Sci-Hub docs only Polygon and Point are accepted: https://scihub.copernicus.eu/twiki/do/view/SciHubUserGuide/OpenSearchAPI#Discover_the_products_over_a_pre

thomasyoung-audet: I've made a quick fix here, but I could make it more of a regex if you want.

valgur: I would not bother with restricting the set of WKT types and would use a pretty general regex pattern instead:
```regex
\s*[A-Za-z]+\s*\(.+\)\s*
```

sentinelsat/scripts/cli.py
```diff
@@ -217,7 +229,24 @@ def cli(
...
 
     if geometry is not None:
-        search_kwargs["area"] = geojson_to_wkt(read_geojson(geometry))
+        # check if the value is an existing path
+        if os.path.exists(geometry):
+            search_kwargs["area"] = geojson_to_wkt(read_geojson(geometry))
+        # check if the value is a GeoJSON
+        else:
+            try:
+                geometry = eval(geometry)
```
valgur, line 238: Using `eval()` is not a good practice in general, since you allow any syntactically valid Python to be executed, which can have a lot of unintended consequences (usually security related, which is not as big of a concern for a CLI utility, but still). Use `json.loads()` instead.
Also, instead of relying on parsing errors to detect non-GeoJSON content, I would check whether `geometry.strip()` starts with `{`. That way any accidental JSON errors would not get ignored.

sentinelsat/sentinel.py
thomasyoung-audet: Great input, thanks. just changed it to `\s*[A-Za-z]*\s*\(.+\)\s*` so it can accept the Point type as it says in the scihub docs.

sentinelsat/scripts/cli.py
thomasyoung-audet: Good call, I'll keep the `eval()` in mind for the future.
I've kept the parsing try except in case there are any errors inside the json string that would make it invalid, but added the check for `{`.

Fernerkundung:
Looks to me like the only thing missing is to adress the failing tests. 

@valgur do you feel fine taking care of the final pass and the merging? I don't want to be the blocker here.

thomasyoung-audet:
Any problems with this or can it merge?

Fernerkundung:
A changelog entry is missing, but everything else looks good from my side 👍 

Anything you want to comment on @valgur?

Fernerkundung:
If no-one has any objections I am going to merge this on Saturday and make it part of a minor release.

sentinelsat/sentinel.py
```diff
@@ -1457,3 +1457,12 @@ def placename_to_wkt(placename):
     footprint = "ENVELOPE({}, {}, {}, {})".format(minX, maxX, maxY, minY)
     placeinfo = feature["properties"]["display_name"]
     return footprint, placeinfo
+
+
+def is_wkt(possible_wkt):
```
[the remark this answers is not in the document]
valgur: Actually, the set of WKT types is so small that they can be added to the validation as well.

```python
def is_wkt(possible_wkt):
    pattern = r"^((MULTI)?(POINT|LINESTRING|POLYGON)|GEOMETRYCOLLECTION|ENVELOPE)\s*\(.+\)$"
    return re.match(pattern, possible_wkt.strip().upper()) is not None
```

sentinelsat/scripts/cli.py
```diff
@@ -69,7 +80,11 @@ def convert(self, value, param, ctx):
     help="End date of the query in the format YYYYMMDD.",
 )
 @click.option(
-    "--geometry", "-g", type=click.Path(exists=True), help="Search area geometry as GeoJSON file."
+    "--geometry",
+    "-g",
+    type=str,
+    help="Search area geometry as GeoJSON file, a GeoJSON string, or a WKT string. "
+    "Valid WKT objects are Polygon and Point.",
```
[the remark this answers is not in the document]
Fernerkundung: This note is directly from the Copernicus Open Access Hub documentation. Seems like this is again a point where the documentation deviates from their own implementation.

Let's got with @valgur comment then and drop this limitation. This will bubble-up as an error should the server drop the support for other WKT types.

sentinelsat/scripts/cli.py
```diff
@@ -217,7 +232,27 @@ def cli(
...
+        if os.path.exists(geometry):
+            search_kwargs["area"] = geojson_to_wkt(read_geojson(geometry))
+        # check if the value is a GeoJSON
+        else:
+            geometry = geometry.strip()
+            if geometry[0] == "{":
+                try:
+                    geometry = json.loads(geometry)
+                    search_kwargs["area"] = geojson_to_wkt(geometry)
+                except SyntaxError:
```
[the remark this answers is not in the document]
thomasyoung-audet: It seems Travis has issues importing `JSONDecodeError` in some of its builds.

sentinelsat/scripts/cli.py
```diff
@@ -237,13 +237,13 @@ def cli(
...
-            if geometry[0] == "{":
+            if geometry.startswith("{"):
                 try:
                     geometry = json.loads(geometry)
                     search_kwargs["area"] = geojson_to_wkt(geometry)
-                except SyntaxError:
-                    logger.info("geometry string starts with '{' but is not a valid GeoJson.")
+                except JSONDecodeError:
+                    logger.error("geometry string starts with '{' but is not a valid GeoJSON.")
+                    exit(1)
```
j08lue, line 246: Just a quick comment: I think we should avoid `print ; exit`, even in the interface, and instead use high-level ways of terminating. Here, perhaps `raise click.UsageError` would be correct?

thomasyoung-audet: Thanks for the input. It felt kind of hacky to me but I wasn't sure how to fix it.

sentinelsat/scripts/cli.py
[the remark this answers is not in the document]
thomasyoung-audet: Its the python 2.7 and mac builds. Any idea how to fix this?

thomasyoung-audet:
Anyone want to take a look at this?

j08lue:
> Anyone want to take a look at this?

You mean help fixing the failing test?

thomasyoung-audet:
I'll be honest the whole time I thought this was a problem with Travis and not something I could fix just playing with the code a bit... Felling a little dumb

Fernerkundung:
Don't worry about it @thomasyoung-audet - we've all been there :smile: 

@valgur can you please re-review the requested changes, then we can get this merged.

Fernerkundung:
Ping @valgur: Have your requested changes been adressed?

I regard the open change request as the final blocker before we can push this over the finish line.

thomasyoung-audet:
Hello @valgur, could we get this merged?? I fixed everything you suggested like 6 months ago

j08lue (approved):
I can confirm that all our change requests have been followed up on and this is ready to merge. Thanks for seeing this through, @thomasyoung-audet.

PR #384, opened 2020-07-08
Improve location formatting, fix cassettes

valgur:
Follow-up for #372.
The location info output was printing `{}` placeholders instead of values and the cassettes were not updated after the final bugfix.
I also tweaked the location info logging further by adding the bounding box extents in km and the box coordinates in a clearer format. So, for example, `sentinelsat --location Metz` prints
```
Querying location: 'Metz, Moselle, Grand Est, France métropolitaine, France' with 9.8 x 8.8 km, 49.060824, 6.136004 to 49.148795, 6.256451 bounding box
```
The box size info is useful since the Nominatim results are not always perfect, i.e. `sentinelsat --location Estonia` yields a bounding box halfway in Poland
```
Querying location: 'Eesti', 807.5 x 1556.7 km, 51.752378, 18.331908 to 65.752378, 32.331908
```

### DIFF
--- a/sentinelsat/scripts/cli.py
+++ b/sentinelsat/scripts/cli.py
@@ -1,4 +1,5 @@
 import logging
+import math
 import os
 
 import click
@@ -212,8 +213,21 @@ def cli(
         search_kwargs.update((x.split("=") for x in query))
 
     if location is not None:
-        wkt, place_name = placename_to_wkt(location)
-        logger.info("The location we are querying is: '%s', '%s'".format(place_name, wkt))
+        wkt, info = placename_to_wkt(location)
+        minX, minY, maxX, maxY = info["bbox"]
+        r = 6371  # average radius, km
+        extent_east = r * math.radians(maxX - minX) * math.cos(math.radians((minY + maxY) / 2))
+        extent_north = r * math.radians(maxY - minY)
+        logger.info(
+            "Querying location: '%s' with %.1f x %.1f km, %f, %f to %f, %f bounding box",
+            info["display_name"],
+            extent_north,
+            extent_east,
+            minY,
+            minX,
+            maxY,
+            maxX,
+        )
         search_kwargs["area"] = wkt
 
     if geometry is not None:
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -1149,15 +1149,15 @@ def geojson_to_wkt(geojson_obj, feature_number=0, decimals=4):
         a GeoJSON object
     feature_number : int, optional
         Feature to extract polygon from (in case of MultiPolygon
-        FeatureCollection), defaults to first Feature
+        FeatureCollection), defaults to first feature
     decimals : int, optional
         Number of decimal figures after point to round coordinate to. Defaults to 4 (about 10
         meters).
 
     Returns
     -------
-    polygon coordinates
-        string of comma separated coordinate tuples (lon, lat) to be used by SentinelAPI
+    str
+        Well-Known Text string corresponding to the GeoJSON geometry.
     """
     if "coordinates" in geojson_obj:
         geometry = geojson_obj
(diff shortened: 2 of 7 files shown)


sentinelsat/scripts/cli.py
```diff
@@ -212,8 +213,21 @@ def cli(
         search_kwargs.update((x.split("=") for x in query))
 
     if location is not None:
-        wkt, place_name = placename_to_wkt(location)
-        logger.info("The location we are querying is: '%s', '%s'".format(place_name, wkt))
+        wkt, info = placename_to_wkt(location)
+        minX, minY, maxX, maxY = info["bbox"]
+        r = 6371  # average radius, km
```
Fernerkundung, line 218: Can you extend the comment how this value is derived and what makes it "average"?

sentinelsat/sentinel.py
```diff
@@ -1149,15 +1149,15 @@ def geojson_to_wkt(geojson_obj, feature_number=0, decimals=4):
...
     decimals : int, optional
         Number of decimal figures after point to round coordinate to. Defaults to 4 (about 10
         meters).
 
     Returns
     -------
-    polygon coordinates
-        string of comma separated coordinate tuples (lon, lat) to be used by SentinelAPI
+    str
+        Well-Known Text string corresponding to the GeoJSON geometry.
```
Fernerkundung, line 1160: Niggle: A WKT is a geometry representation that in itself is not related to GeoJSON, except that they both refer to spatial geometries. The phrasing sounds like this is somehow a WKT+GeoJSON string.

sentinelsat/sentinel.py
```diff
@@ -1427,33 +1427,42 @@ def _parse_odata_response(product):
...
+    ------
+    ValueError
+        If no matches to the place name were found.
+
     Returns
     -------
-    full name and coordinates of the queried placename
-        list in the form [wkt string in form 'ENVELOPE(minX, maxX, maxY, minY)', string placeinfo]
+    wkt_envelope : str
+        Bounding box of the location as an 'ENVELOPE(minX, maxX, maxY, minY)' WKT string.
```
Fernerkundung, line 1447: A bounding box and an evenlope *can* be identical but don't have to be. Can we use a single terminology throughout the codebase?

The Envelope vs. Extent definitions can be rather confusing, see for instance PostGIS [ST_Envelope](https://postgis.net/docs/ST_Envelope.html) vs. [ST_Extent](https://postgis.net/docs/ST_Extent.html) (which I find to be a bad pattern).

sentinelsat/sentinel.py
```diff
@@ -1427,33 +1427,42 @@ def _parse_odata_response(product):
...
     )
-    # Check that the response from Openstreetmapserver has status code 2xx
-    # and that the response is valid JSON.
     rqst.raise_for_status()
-    jsonlist = rqst.json()
-    if len(jsonlist["features"]) == 0:
-        raise ValueError('Unable to find a matching location for "{}"'.format(placename))
+    features = rqst.json()["features"]
+    if len(features) == 0:
+        raise ValueError('Unable to find a matching location for "{}"'.format(place_name))
```
Fernerkundung, line 1459: This would also be triggered if the nominatim endpoint is unavailable, wouldn't it? In that case the error message might be slightly confusing.

sentinelsat/sentinel.py
```diff
@@ -1427,33 +1427,42 @@ def _parse_odata_response(product):
...
     # Get the First result's bounding box and description.
-    feature = jsonlist["features"][0]
+    feature = features[0]
     minX, minY, maxX, maxY = feature["bbox"]
-    footprint = "ENVELOPE({}, {}, {}, {})".format(minX, maxX, maxY, minY)
-    placeinfo = feature["properties"]["display_name"]
-    return footprint, placeinfo
+    # ENVELOPE is a non-standard WKT format supported by Solr
+    # https://lucene.apache.org/solr/guide/6_6/spatial-search.html#SpatialSearch-BBoxField
```
Fernerkundung, line 1464: ❤️

sentinelsat/scripts/cli.py
valgur: At this level of precision (which is more than enough for this simple use case) pretty much all ways of computing the average radius of the Earth round to that value. https://en.wikipedia.org/wiki/Earth_radius

sentinelsat/sentinel.py
valgur: Good point. Is "Well-Known Text string representation of the geometry" better?

sentinelsat/sentinel.py
valgur: I guess there can be, but I don't see any way they could have a different meanings in our use cases?
Either way, I personally prefer "bounding box" as it sounds a bit clearer to me.

sentinelsat/sentinel.py
valgur: It should generate a HTTP error from `rqst.raise_for_status()` if Nominatim is not available.

sentinelsat/sentinel.py
Fernerkundung: Sounds good :+1:

sentinelsat/sentinel.py
Fernerkundung: Let's go with that then :+1:

Fernerkundung:
LGTM!

If you add the changes to the changelog in this or another of your PRs I think we can get this merged right away @valgur :+1:

Issue #383, opened 2020-07-08
SciHub requiring keyword values to be within quotes


radosuav:
It appears that there was a subtle change to SciHub interface recently. Previously I could specify extra keywords using single quotes, e.g.:
```
other_search_keywords["orbitdirection"] = "descending"
other_search_keywords["timeliness"] = "Near Real Time"
```
Now they have to be in double quotes (i.e. ' followed by "):
```
other_search_keywords["orbitdirection"] = '"descending"'
other_search_keywords["timeliness"] = '"Near Real Time"'
```
Otherwise apihub never responds to the query ("timeliness") or ignores the extra keyword ("orbitdirection").

There are some keywords which still work with single quotes, e.g. 
`other_search_keywords["productlevel"] = "L2"`
seems to work fine.

To be on the safe side, maybe all keywords should be put within extra quotes.

valgur:
Interesting. We already take care to handle spaces in values correctly (#169), but something might have changed on the server side. I will try to take a closer look at this soon.

valgur:
Fixed by #390.

PR #390, opened 2020-07-24
Fix space handling, improve set input performance, improve --uuid and --name args

### LINKED ISSUE #387: --name overrides -s and -e?
Hi all, hopefully I can get some help with this.

I'm trying to download T32PLT Sentinel 2 images using sentinelsat for the period of 04/03/2019 and 27/02/2020. I'm using the code underneath:

`sentinelsat -u username -p password --start '20190304' --end '20200227' --name '*_T32PLT_*' --sentinel 2 --download --path '/path/to/folder'`

Unfortunately this seems to download every single T32PLT image to have ever existed in the archive, and the --start and --end don't seem to affect this at all. Having the same issue with Sentinel 1 by specifying:

 `--name 'S1A_IW_SLC*'`

Any suggestions are appreciated.

valgur:
* Fixes the server-side space handling change issue reported in #383 by quoting instead of backslash-escaping the strings.
* Uses `(key:value OR key:value)` format instead of the previous `key:(value OR value)` format, which for some reason takes tens of seconds per query.
* Got rid of the special handling of `--uuid` and `--name` CLI arguments. The product IDs are now simply passed to `api.query()` as a set. They no longer also ignore the date range arguments (fixes #387).
* Empty queries raise a `ValueError` immediately on the client side instead of letting the server generate it.
* Added stricter checks for empty keyword values in queries, which would cause server-side errors.
* Updated all cassettes and made some tests less dependent on specific products.

TODO
- [x] add changelog entries
- [ ] merge other PRs first


valgur:
I rebased these changes onto `main`. Can be merged after #450.

j08lue (approved):
Great work. Query formatting logic got a lot simpler and good idea to move it to a separate method. 👍

PR #404, opened 2020-09-28
Added feature to download quicklooks directly

### LINKED ISSUE #361: Adding functionality for downloading quicklooks
In some cases, it is good to evaluate the quicklooks before proceeding to download all the images. This can help reduce the number of images ultimately downloaded.
As such, I'd like to request the functionality of a small flag in the command-line that sets `sentinelsat` to download the quicklooks, rather than the actual files. I propose `--quicklooks` and `-ql`.

mackland:
Fixes #361 

A user can provide --quicklooks to download quicklooks directly without having to download full product.

- Kept this separate from download_all so that it can be run with or without -d option, this way quicklooks can be inspected as download is running or without download to get an idea how sensing looks.
- This is done without the retry functionality as it is quick enough to not need it.
- Added error handling for missing quicklooks and incorrect content-type. These are logged similar to how downloaded products logged in a .txt file in _path_


mackland:
Updated to take backwards compatibility into consideration.

black would change much more than just the code submitted, am I missing some settings? Nothing in contribute.rst and couldnt find anything in .github/ (I just ran _black ._ as stated)

Fernerkundung:
> black would change much more than just the code submitted, am I missing some settings? Nothing in contribute.rst and couldnt find anything in .github/ (I just ran black . as stated)

Can you try again with merging the latest changes in master? Looks like 3 files indeed needed to be formatted to adhere to the latest version of black.

sentinelsat/sentinel.py
```diff
@@ -909,6 +910,54 @@ def _download_online_retry(
...
+        Returns
+        -------
+        dict[string, dict]
+            A dictionary containing the error of products where either
+            quicklook was not available or it had an unexpected content type
+        """
+
+        self.logger.info("Will download %d quicklooks", len(products))
+
+        outdir = Path(directory_path + "/quicklooks")
```
Fernerkundung, line 938: Please use pathlib to construct the path instead of string concatenation. This ensures that it will work on all OS's.

sentinelsat/sentinel.py
```diff
@@ -909,6 +910,54 @@ def _download_online_retry(
 
         return ret_val
 
+    def download_quicklooks(self, products, directory_path="."):
```
Fernerkundung, line 913: Please provide a test for this new function.

Fernerkundung (changes requested):
Thank you for your contribution!

This is looking promising. I requested some small changes.

I haven't inspected the logic of the new function throroughly but it seems to use a different pattern to our existing functions. Do you mind giving this a second look @j08lue ?

sentinelsat/sentinel.py
mackland: I changed this to use os.path since pathlib isnt in the standard library for older versions. In updated version I do avoid using string concatenation. Will commit together with other changes highlighted on here. Thanks for having a look!

sentinelsat/sentinel.py
```diff
@@ -910,6 +910,74 @@ def _download_online_retry(
...
+        Returns
+        -------
+        dict[string, dict]
+            A dictionary containing the error of products where either
+            quicklook was not available or it had an unexpected content type
+        """
+
+        self.logger.info("Will download %d quicklooks", len(products))
+
+        outdir = join(directory_path, "quicklooks")
```
j08lue, line 940: Please remove the outdir (sub-directory) logic from the core function. Which ever directory the user passes to `download_quicklooks` should be the one they are written to. This might end up being the same directory the data files are downloaded to, but if a user needs to change that, they can use the Python API.

sentinelsat/scripts/cli.py
```diff
@@ -255,6 +261,14 @@ def cli(
         with open(os.path.join(path, "search_footprints.geojson"), "w") as outfile:
             outfile.write(gj.dumps(footprints_geojson))
 
+    if quicklook:
+        failed_quicklooks = api.download_quicklooks(products, path)
+        if failed_quicklooks:
```
j08lue, line 266: Please do not write a text file with failed quicklooks. Just write a message to the terminal or so. Fine for the function to return this information, though. Maybe something like `f"Some quicklooks failed to download: {len(products)} out of {len(failed_quicklooks)}."`. If someone needs logic about what to do in case of failing quicklooks, they need to use the Python API.

j08lue (changes requested):
This is pretty neat. Only thing to change is to remove some of the custom logic, like creating a subdirectory for quicklooks and writing something about missing quicklooks to a file. If a user wants this kind of logic, they can code that up themselves.

And we need a test, please.

sentinelsat/scripts/cli.py
mackland: Thanks, will change this and log to screen instead. Had a look at writing tests over the weekend which made me rewrite so the functions take parameters similar to download(), download_all() and then can test them similar to what you are already doing there.

Will try and get that commit up tomorrow, appreciate you guys having a look at this.

j08lue (approved):

sentinelsat/sentinel.py
```diff
@@ -1354,6 +1453,13 @@ def _parse_odata_timestamp(in_date):
     return datetime.utcfromtimestamp(seconds) + timedelta(milliseconds=ms)
 
 
+def _get_quicklook_url(id):
+    url = "https://scihub.copernicus.eu/apihub/odata/v1/Products('{id}')/Products('Quicklook')/$value".format(
+        id=id
+    )
```
Fernerkundung, line 1459: This will fail if the user provided a different SciHub URL, e.g. https://apihub
.copernicus.eu

Please re-use the url attribute of the api class, which is where we handle the SciHub base-url logic.

Fernerkundung (changes requested):
Thanks for adding the thorough tests.

This is looking very good to push over the finish line 👍 

Two minor requests:
1. The quicklook URL should be created from the api base URL provided by the user.
2. Please add your changes to the CHANGELOG.rst

sentinelsat/sentinel.py
mackland: Thanks for catching that, updated to use self.api_url and added the quicklook feature to changelog as well

mackland:
Are you happy with the latest changes? I wasnt following this closely but think it can be merged now unless you want anything added? @Fernerkundung @j08lue

Fernerkundung (approved):
Looks good. Thanks for contributing this feature to sentinelsat 👍

PR #407, opened 2020-10-13
Changed the functioning of the '--footprints' option.

### LINKED ISSUE #153: --footprints parameter
## Question
**Should we change the `--footprints` parameter's behavior to also include the name of the file that will be created?**

Let's make it a poll. I encourage anyone to leave a thumbs up or down to voice your opinion. Comments are welcome too, of course.

## Reasoning
I've been thinking that the `--footprints` flag would be more useful if it also took the name/path of the file to be created instead of always writing it to `search_footprints.geojson`.

An example use case might be:
```
sentinelsat -g germany_aoi.geojson --sentinel 1 --footprints s1_germany_footprints.geojson
```
On the other hand, forcing the user to always provide the destination filename is a minor inconvenience as well (which we could mitigate a bit by automatically appending the `.geojson` suffix if it is missing).

Since I rarely use the `--footprints` feature myself (I rely on the Python API) and the choice is not obvious, I would like to hear everyone else's take on this.

IpsumCapra:
<h1>Changed the way the '--footprints' option works.</h1>
The footprints option can now take a relative directory and option other name. This relative directory (together with optional custom name) will be joined with the path option.

Closes #153.

### DIFF
--- a/sentinelsat/scripts/cli.py
+++ b/sentinelsat/scripts/cli.py
@@ -136,9 +136,9 @@ class CommaSeparatedString(click.ParamType):
 )
 @click.option(
     "--footprints",
-    is_flag=True,
-    help="""Create a geojson file search_footprints.geojson with footprints
-    and metadata of the returned products.
+    default=None,
+    help="""Create a geojson file with default name search_footprints.geojson with footprints
+    and metadata of the returned products, at specified path (and optional custom name).
     """,
 )
 @click.option("--info", is_flag=True, is_eager=True, help="Displays the DHuS version used")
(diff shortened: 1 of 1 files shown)


IpsumCapra:
Changing the way --footprints works seems to have broken the test, which was to be expected, I guess.

Fernerkundung:
If the tests require changing to handle your new feature please also change the tests.

sentinelsat/scripts/cli.py
```diff
@@ -250,9 +250,13 @@ def cli(
         end = end or "NOW"
         products = api.query(date=(start, end), order_by=order_by, limit=limit, **search_kwargs)
 
-    if footprints is True:
+    if footprints is not None:
+        if os.path.isdir(footprints) or footprints[len(footprints) - 1] == os.path.sep:
```
Fernerkundung, line 254: `footprints[len(footprints) - 1] == os.path.sep` this would fail if a user does not provide a trailing slash.

works: `/some/path/`
fails: `/some/other/path`

Fernerkundung (changes requested):
This is a good start to the functionality described in the issue. Please add tests to test the new functionality you introduced to the footprints parameter.

sentinelsat/scripts/cli.py
IpsumCapra: That's what `os.path.isdir(footprints)` is for. The other part is for checking whether the user is specifying a directory that does not exist. If the trailing slash is left out, it will use that part as the file to write to. Does that clear things up?

j08lue: I don't think we need special treatment for trailing slashes in paths? `os.path.join` treats both the same and so does `os.path.isdir`. So you can just remove that `or` condition.

IpsumCapra: I added it for non-existent directories to be added, but I'll remove it, then.

valgur:
Thanks for the PR, @IpsumCapra! I extended and merged this feature in #462.

PR #412, opened 2020-12-22
Add main script

avalentino:
Add a ``__main__.py`` file that simply calls ``sentinelsat.scripts.cli.cli`` to be able to run sentinelsat CLI as follows:
```
$ python3 -m sentinelsat --help
```

This could be handy when one uses multiple python environments.

### DIFF
--- a/sentinelsat/__main__.py
+++ b/sentinelsat/__main__.py
@@ -1,0 +1,5 @@
+import sys
+
+from sentinelsat.scripts.cli import cli as main
+
+sys.exit(main())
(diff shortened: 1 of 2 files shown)


sentinelsat/__main__.py
```diff
@@ -0,0 +1,5 @@
+import sys
+
+from sentinelsat.scripts.cli import cli as main
```
j08lue, line 3: Why import as `main` and not leave as `cli`?

j08lue (approved):
Cool, thanks. Fine with me like this.

sentinelsat/__main__.py
avalentino: Just a matter of taste, it seems to me more clear.
But I can change if you want.

j08lue: No, thanks, never mind. 😄

avalentino:
Thanks @j08lue

j08lue:
Thank you for your contribution, @avalentino! I forgot we also need to add this change to our Changelog. Could I ask you to open another PR for that?

avalentino:
Hi @j08lue there is already a line in CHANGELOG.rst documenting this PR.
Do you mean something else?

j08lue:
Haha, sorry, brain still in holiday mode... All good then.

Issue #386, opened 2020-07-10
LTA request out of memory


frx08:
It is possible that the product odata gives status offline but the product is available.
What happen is that the `_trigger_offline_retrieval` is called and instead of just triggering the retrieval the product is actually downloaded 
https://github.com/sentinelsat/sentinelsat/blob/4bf367a992a9e29a9fa10a56881d8519dd755af0/sentinelsat/sentinel.py#L658
since this call is not ready to handle huge downloads the memory grows and in this case my process is killed

Fernerkundung:
I assume this is limited to the endpoint `https://scihub.copernicus.eu/dhus` or did you test this with multiple endpoints that support LTA?

frx08:
yes my case is for `https://scihub.copernicus.eu/dhus` and i assume it is only for it. in this days copernicus server are badly behaving so this situation poped out
for the moment my dirty workaround is to do this:
`
def _disabled_trigger_offline_retrieval(self, url):
    _msg = 'trigger_offline_retrieval is disabled'
    self.logger.error(_msg)
    raise SentinelAPILTAError("trigger_offline_retrieval is disabled", type('Response', (object,), { 'status_code': 500, 'reason':_msg }))
SentinelAPI._trigger_offline_retrieval = _disabled_trigger_offline_retrieval
`

valgur:
Thanks for reprting this. That's an oversight for sure.
Adding `stream=True` to the GET request or replacing GET with HEAD should probably fix this.

LLeiSong:
Same problem happened to me. Not totally sure, but there is another related problem: when the imagery go back to offline, but the product info still have `Online: True`. This also cause problem.

PR #421, opened 2021-01-21
WIP: Use HEAD request for LTA triggers

lucadelu:
working on #386, replaced get with head as suggested by @valgur

The option stream=true doesn't work for me


valgur (changes requested):
Please update the `test_trigger_lta_success` and `test_trigger_lta_failed` tests as well. Otherwise LGTM.

valgur:
Thanks, @lucadelu! I extended and continued the PR in #457.

PR #423, opened 2021-02-08
Drop python2 (Closes  #416)

### LINKED ISSUE #416: Drop Python 2 support?
Should we finally drop Python 2 support? It has been very little effort to keep it until now, but causes some trouble for example in #414 and I would say it has long been time to say goodbye...

avalentino:
This PR drops support to Python 2 and modernize a little bit the code (the ``pyupgrade`` tool has been used for that).
Also the ``six`` library is no longer necessary of course. It has been removed form dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,11 +24,11 @@ branches:
   only:
   - master
 before_install:
-- python -m pip install -U pip
+- pip3 install -U pip
 install:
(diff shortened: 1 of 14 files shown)


j08lue (approved):
Nice! We did not take the decision yet to drop 2.7, but if we do, this can be how.

Issue #396, opened 2020-08-05
Expose fileobj api for downloading and streaming files


daniel-j-h:
Hey there :wave: amazing project and thanks for what you are doing! :tada: :hugs: 

I have a small suggestion / some feedback. Right now the download functions take a directory and then will write a file there.

Would it be possible to expose an api taking a file-like object for writing so that users could e.g. pass `io.BytesIO` and don't have to go through the filesystem? Would it then also be possible to use this for streaming the data e.g. from the download function directly to a file store like s3? Otherwise users have to download gigabytes of files to disk, then upload them to a file store.

And while we're at it, is it possible to download partial files from the server? For example say I only want a specific band from Sentinel-2 for the last two years. Right now I could only find ways to download all the data at once and then filter locally which is quite wasteful in terms of disk space, bandwith, etc. This might be a server limitation though?

:satellite: :milky_way:

j08lue:
Thanks a lot for the 💐 and the suggestion.

The file-on-disk logic makes up quite a large part of `api.download` and [`api._download`](https://github.com/sentinelsat/sentinelsat/blob/4bf367a992a9e29a9fa10a56881d8519dd755af0/sentinelsat/sentinel.py#L1095-L1125
) and IMO it would be overkill to make this configurable / replaceable. You could perhaps subclass `SentinelAPI` and change the two methods to your needs. But I think it will be easiest for you to just take the results from the query and process the download links as you want - stream them or copy them to S3 or so.

But if you can prove me wrong and propose an elegant way of exposing a fileobj, we are always open for PRs. ;-)

Please write one question per issue. But your second question already has one here: #291.

daniel-j-h:
Gotcha! Thanks for the explanation, looking at the code it's more than just putting bytes into a buffer :thinking: Handling the download links from the query on my own is something I haven't thought of :ok_hand: I think we're good to close here, all good! :hugs:

PR #430, opened 2021-04-06
Add streaming option

fwfichtner:
As discussed in https://github.com/sentinelsat/sentinelsat/issues/396 and https://github.com/dlr-eoc/ukis-pysat/pull/132 it could be a useful addition to allow the user to get a `requests` response object ready for streaming with a new method.

I opted not to handle any LTA triggers for now. For the test I had to adapt the checksum manually, because the file I got had another one. That's weird, but I did not want to adapt other tests or recordings as well.

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -1253,7 +1253,7 @@ class SentinelAPI:
         product_info = self.get_product_odata(id)
         if not product_info["Online"]:
             raise NotImplementedError("Product is offline, no retrieval implemented.")
-        r = requests.get(
+        r = self.session.get(
             product_info["url"],
             stream=True,
             auth=self.session.auth,
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -1237,6 +1237,29 @@ class SentinelAPI:
         )
         product["quicklook_url"] = url
 
+    def get_stream(self, id):
+        """Exposes requests response ready to stream product to e.g. S3.
+
+        Parameters
+        ----------
+        id : string
+            UUID of the product, e.g. 'a8dd0cfd-613e-45ce-868c-d79177b916ed'
+
+        Returns
+        -------
+        tuple:
+            raw socket response from server, contains product's info as returned by get_product_info()
+        """
+        product_info = self.get_product_odata(id)
+        if not product_info["Online"]:
+            raise NotImplementedError("Product is offline, no retrieval implemented.")
+        r = requests.get(
+            product_info["url"],
+            stream=True,
+            auth=self.session.auth,
+        )
+        return r.raw, product_info
+
 
 def read_geojson(geojson_file):
     """Read a GeoJSON file into a GeoJSON object."""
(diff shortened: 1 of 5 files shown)


sentinelsat/sentinel.py
```diff
@@ -1237,6 +1237,29 @@ def _add_quicklook_url(self, product):
...
+
+        Returns
+        -------
+        tuple:
+            raw socket response from server, contains product's info as returned by get_product_info()
+        """
+        product_info = self.get_product_odata(id)
+        if not product_info["Online"]:
+            raise NotImplementedError("Product is offline, no retrieval implemented.")
+        r = requests.get(
```
j08lue, line 1256: All other places, we do this:

```suggestion
        r = self.session.get(
```

Could we follow that pattern here, or is there a reason for using `requests` directly?

j08lue:
>  For the test I had to adapt the checksum manually, because the file I got had another one. That's weird, but I did not want to adapt other tests or recordings as well.

So the checksum from the metadata (`product_info["md5"].lower()`) is different from that you compute from the file you created (`expected_path.computehash("md5")`), so you adjusted the `product_info` manually? This could cause a lot of head scratching when we re-record the cassette at some point... And besides that, we should get to the origin of this, so no manual patching is necessary and we know what is going on.

So there are three possibilities:
1. the checksum in the metadata is wrong. But that would mean running the other download tests without cassettes should fail, too.
2. the way you write the file from stream causes the file to be different from a direct download to disk
3. the way you compute the checksum is different than expected.

To rule out option 3, maybe you should use `SentinelAPI._md5_compare` instead of `path.computehash("md5")`. We use `computehash` elsewhere, but that is actually to compare a successfully downloaded file with a tinkered one, not file against metadata.

Also check option 1, please, by running `pytest --disable-vcr tests/test_download::test_download` and making sure that it does not already fail at https://github.com/sentinelsat/sentinelsat/blob/faa745f7e0c3bb7512ff69edbdd8e572c4fe39c9/tests/test_download.py#L96.

fwfichtner:
I am either overseeing soemthing or getting another checksum from the server for this product as the one saved in the recordnings (1). However, it shouldn't matter because when only the recordings are used it fits together. That's why I manually adapted the checksum in [here](https://github.com/sentinelsat/sentinelsat/pull/430/files#diff-008e10a58587db6a56d8e749165694c842cf0e2b42e3811ec96c32633341058fR17). When running without cassettes it works as well, because it compares against the new response from the server. 
If my obersavtion is true the recordings should be updated, but I would consider this to be outside the scope of this issue.

I also worked in your recommendations, thanks for that. I tried several options for (2), but the checksum stays the same.

j08lue:
> If my obersavtion is true the recordings should be updated, but I would consider this to be outside the scope of this issue.

It would be great if you could include the new cassettes in this PR, please. Otherwise we just add that to some backlog that we might or might not get to... All it should take is to run your test and record a new VCR cassette `pytest --vcr-record=new_episodes tests/test_download::test_get_stream`. If the cassette for retrieving the `smallest_online_products` is re-recorded, you can just run the whole test suite afterwards and see whether the change affected any of the other tests. Hopefully not. 😄

If you cannot be persuaded to do this, I am fine with merging and opening a follow-up issue, but we need to give it priority then...

j08lue:
OK. I'll merge this PR as it is then and open a backlog issue to re-record cassettes and fix issues.

j08lue:
The only thing missing would be a CHANGELOG entry, please, and adding your name to the AUTHORS list, if you like.

fwfichtner:
> 
> 
> The only thing missing would be a CHANGELOG entry, please, and adding your name to the AUTHORS list, if you like.

Done with cea8130 . Thanks for the feedback and your work with `sentinelsat`.

PR #435, opened 2021-04-10
Recommend apihub

j08lue:
As mentioned in many issues, the recommended SciHub endpoint for machine access is `/apihub`, NOT `/dhus`.

For a while `/dhus` was a convenient drop-in replacement when user credential changes (including new sign-ups) had not propagated to `/apihub` yet. But the Copernicus Open Access Hub now prevents that by heavily rate-limiting access to `/dhus`.

In short, we should not anywhere recommend that users call `/apihub` on the Copernicus Open Access Hub anymore.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -77,7 +77,7 @@ Python Library
   from datetime import date
 
   # connect to the API
-  api = SentinelAPI('user', 'password', 'https://scihub.copernicus.eu/dhus')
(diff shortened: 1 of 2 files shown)


j08lue:
I was hoping to see some Travis CI action on this PR, but it seems like we have run out of credits for this period. Our pretty big build matrix is perhaps a bit excessive? But credits should get replenished today. I'll merge this anyways, as it is just docs.

PR #448, opened 2021-04-30
Gracefully handle cancelled futures

avalentino:


sentinelsat/sentinel.py
```diff
@@ -791,15 +791,16 @@ def download_all(
...
-                    product_info = dl_task.result()
-                    downloaded_prods[product_info["id"]] = product_info
-                # This elif catches the first dl_task that did not complete because
-                # the product was not online and _download_online_retry returned None
-                elif not dl_task.exception() and dl_task.result() is None:
-                    stop_event.set()
-                    for task in dl_tasks:
-                        task.cancel()
+                if not dl_task.cancelled() and not dl_task.exception():
+                    if dl_task.result():
```
j08lue, line 795: Perhaps make this

```suggestion
                    if dl_task.result() is not None:
```

and then just an `else` below? This would be more robust towards results that do not easily evaluate to a boolean. These should not occur here, but still...

Also, you could consider assigning the result to a variable, rather than calling `.result` several times. Not sure whether the return value of `.result` could change between the calls and then you get weird behavior.

j08lue (commented):
Great you are looking into this issue. I am not fully on top of the behavior here, just put in a few comments regarding the evaluation and state of the result.

sentinelsat/sentinel.py
avalentino: Done, thanks, in this way is much clearer.

valgur (approved):
LGTM

j08lue (changes requested):
Yes, thanks! Only missing a CHANGELOG entry, no?

avalentino:
> Yes, thanks! Only missing a CHANGELOG entry, no?

@j08lue done

j08lue (approved):

PR #438, opened 2021-04-11
Migrate to GHA (Closes #437)

avalentino:
Set-up CI on GHA and drop Travis-CI.

Tested Python versions: 3.6-3.9 (please let me know if you steel want to test Python 3.5).
Notifications are not active (commented code). You need to set the Gitter token in secrets for that.

Closes #437.

PR #449, opened 2021-04-30
Fix product node issues on Windows

valgur:
Fixes the test failures that cropped up in #438.

- Mark downloaded data as 'binary' in git to prevent the inserted line break characters on Windows from messing up cassette playback. The downloaded files used in test are now stored in `fixtures/vcr_cassettes/data` to make them easier to distinguish.
- Make `/` the path separator for within-product files irregardless of the OS path separator. I think that's more convenient and expected behavior for the user. Do you agree @avalentino?
- Some minor code tidying.


sentinelsat/products.py
```diff
@@ -86,28 +86,25 @@ def nodefilter(node_info: dict) -> bool:
...
         elif urltype is None:
-            data["urltype"] = ""
-        else:
-            data["urltype"] = urltype
-        return "{api_url}odata/v1/Products('{id}')/Nodes('{title}.SAFE')/{path}{urltype}".format(
-            **data
-        )
+            urltype = ""
+        # else: pass urltype as is
+        return f"{self.api_url}odata/v1/Products('{id}')/Nodes('{title}.SAFE')/{path}{urltype}"
```
avalentino, line 101: +1, but you need to drop support Python 3.5 to use it. In ``setap.py`` 3.5 is still listed.

avalentino (approved):
LGTM

avalentino:
> Make / the path separator for within-product files irregardless of the OS path separator. I think that's more convenient and expected behavior for the user. Do you agree @avalentino?

@valgur absolutely yes :)

j08lue (approved):
Nice fixes and improvements.

PR #445, opened 2021-04-29
Change of API Hub URL (28 April 2021) 

MrChebur:
https://scihub.copernicus.eu/news/News00868

```
Open Access Hub users are informed that the API Hub URL has changed from 
https://scihub.copernicus.eu/apihub/ to 
https://apihub.copernicus.eu/apihub/ as of 28 April 2021.
The former URL will not be dismissed. An automatic redirection to the new URL is performed transparently. 
```

PR #450, opened 2021-05-02
API Hub URL change, update cassettes, fix regressions

### LINKED ISSUE #431: Re-record VCR cassettes and fix breaking tests
Seems like we should soon do this again:

> Unfortunately many other tests are affected once I change that file. Tried to fix them one by one, but `test_download_all_quicklooks_one_fail` keeps failing and I cannot get my head around why that is at the moment. 
> 
> Simply re-recording everything lead to other exceptions.

_Originally posted by @fwfichtner in https://github.com/sentinelsat/sentinelsat/issues/430#issuecomment-815029229_

valgur:
Updated the URL as mentioned in #445 but applied it to all of the codebase. Since this affects all cassettes, I also re-recorded all of them (as is annual tradition). This highlighted a couple of server-side changes that needed some code updates on our side:

- Server-side error info was seriously lacking previously and was not even raised as errors in the case of syntax errors, for example. This has improved considerably and we no longer need to resort to hack-y logic to catch these.
- Max query length changed by 1%.
- The Sentinel-2 product summary info displayed on CLI changed a bit.

This includes changes from #449, which has to be merged first.
Fixes #431.


j08lue (approved):
Great work!

PR #452, opened 2021-05-03
Improve code coverage, various fixes

valgur:
- Refactored download code a bit to get rid of duplication in `SentinelProductsAPI`.
- Replaced `os` calls with `pathlib.Path` everywhere.
- `self.nodefilter` would not ever get used currently and is not initialized correctly, so removed it.
- Simplified node filters a bit.
- Improved test coverage for the custom exceptions. This actually highlighted some cases that needed improvement, such as broken Unicode content with the old logic.


sentinelsat/sentinel.py
```diff
@@ -626,22 +632,19 @@ def download(self, id, directory_path=".", checksum=True, **kwargs):
...
         if not skip_download:
             # Store the number of downloaded bytes for unit tests
+            temp_path.parent.mkdir(parents=True, exist_ok=True)
             product_info["downloaded_bytes"] = self._download(
                 product_info["url"], temp_path, self.session, product_info["size"]
             )
-
         # Check integrity with MD5 checksum
-        if checksum is True:
+        if verify_checksum is True:
```
j08lue, line 642:
```suggestion
        if verify_checksum:
```

sentinelsat/sentinel.py
```diff
@@ -595,29 +595,35 @@ def download(self, id, directory_path=".", checksum=True, **kwargs):
...
-                    getsize(temp_path),
+                    size,
                     product_info["size"],
                 )
-                remove(temp_path)
-            elif getsize(temp_path) == product_info["size"]:
-                if checksum is True and not self._md5_compare(temp_path, product_info["md5"]):
+                temp_path.unlink()
+            elif size == product_info["size"]:
+                if verify_checksum is True and not self._md5_compare(
```
j08lue, line 617:
```suggestion
                if verify_checksum and not self._md5_compare(
```

sentinelsat/products.py
```diff
@@ -1,32 +1,10 @@
-import os
-import shutil
+from fnmatch import fnmatch
```
j08lue, line 1: The Google Style Guide [discourages](https://google.github.io/styleguide/pyguide.html#22-imports) any `from` imports of functions and classes, except for `typing`.

We do not follow the G style guide explicitly, but should we perhaps still adhere to this nice rule? That would require you to use `fnmatch.fnmatch` and `pathlib.Path` everywhere.

j08lue (changes requested):

j08lue:
You covered a lot of ground here and made the code a lot nicer, again breaking up functionality into more logical units / functions and then all nice nice Python >= 3.6 improvements... 💄 Only one style considerations, apart from the two tiny fixes I'll merge (weeding out `is True`).

sentinelsat/products.py
valgur: I somewhat agree with that style guide suggestion, but within reason. I find the `pathlib.Path` to be such a basic and common class in Python that using it as `from pathlib import Path` seems almost idiomatic and the alternative looks clunky to me. I agree for the `fnmatch.fnmatch`, though.

valgur:
Also, @avalentino, you are welcome to review the code here and elsewhere as well. Especially here as it touches quite a bit of code authored by you. I can't mark you as a requested reviewer since you are not part of the sentinelsat org.

sentinelsat/sentinel.py
```diff
@@ -1183,41 +1183,44 @@ def name_from_path(path):
 
             # Collect
             if name not in names_from_paths:
-                paths.append(join(directory, name + ".zip"))
+                paths.append(directory / (name + ".zip"))
```
avalentino, line 1186: Not sure, should we use ``_get_filename`` here?

sentinelsat/sentinel.py
```diff
@@ -1423,18 +1426,18 @@ def _check_scihub_response(response, test_json=True, query_string=None):
     except (requests.HTTPError, ValueError):
         msg = None
         try:
-            msg = response.headers["cause-message"]
-        except Exception:
+            msg = response.json()["error"]["message"]["value"]
+        except:
```
avalentino, line 1430: Why you have removed ``Exception`` here (and in the following)?
Do you also want to catch e.g. ``KeyboardInterrupt`` and things like that?

avalentino (approved):
Very good job.
Just a couple of minor notes (inline).

sentinelsat/sentinel.py
valgur: Oh, wow, I've always just assumed that plain `except` is equivalent to `except Exception` rather than `except BaseException` and was not aware that `KeyboardInterrupt` derives from `BaseException`. That's super useful to know. Thank you! I'll fix that for sure.

sentinelsat/sentinel.py
valgur: Good catch. Thanks! Fixed it.

j08lue (approved):

PR #453, opened 2021-05-03
Run weekly live tests without cassettes in CI

valgur:
The pre-recorded cassettes we have been using are great for general testing but they have the downside of missing any server-side changes that do occasionally happen. This PR sets up a weekly (every Sunday at 00:00) automated test-run without cassettes. If the tests fail, the workflow automatically creates a new issue, [like this](https://github.com/valgur/sentinelsat/issues/4). The newly recorded cassettes are also made available as a build artifact.
Here's an example of a successful run: https://github.com/valgur/sentinelsat/actions/runs/807920568

I also tweaked the test configuration to reduce the most time-consuming parts of a live test-run from about 5 mins to 10 seconds or so. The tests seemed to hang in the CI otherwise for some reason.


j08lue (approved):
Great you automated the cassette scrubbing and update part!

This could cause a lot of issues we have to investigate (once a week, at least), and sometimes may be due to temporary SciHub outages, which are not uncommon. But let us just try this, we can always adjust later.

valgur:
> This could cause a lot of issues we have to investigate (once a week, at least), and sometimes may be due to temporary SciHub outages, which are not uncommon. But let us just try this, we can always adjust later.

That's a good point. I added a server health check before the live test run, which will cause the live test to be skipped and regular testing to be run. The latter is still useful to catch any regressions due to dependencies or other reasons.

j08lue (approved):
Nice improvement with the branch based on the result of the API health check. 👍

PR #457, opened 2021-05-08
Use HEAD for online status checks, fix inconsistent application of timeout param

### LINKED ISSUE #386: LTA request out of memory

valgur:
Fixes #386 and extends the #421 PR. 

I also noticed that the `self.timeout` parameter was used somewhat inconsistently across queries and fixed that by making it a default session parameter. Setting `auth=self.session.auth` for all queries no longer seems necessary either - all tests pass without it and the CLI works as expected.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -51,11 +51,13 @@ Deprecated
 
 Fixed
 ~~~~~
-* fix location information for Nominatim bounding box queries (#384)
+* Fix location information for Nominatim bounding box queries (#384)
 * Get file name extension more reliably from either header or internal logic (in particular for S5 products #270) (#378 @valgur)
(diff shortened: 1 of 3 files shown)


j08lue (approved):
Vital spring cleaning - cleaner code and fix of inconsistencies. 👍 👍

PR #461, opened 2021-05-08
Fix broken LTA test

valgur:
- Fixed fetching of the list of small archived products.
- Fixed `test_download_all_lta` that has been causing issues.
- Adding some more scrubbing of cassette contents to minimize spurious changes due to varying returned product counts.


j08lue (approved):
🧹 👍

Issue #391, opened 2020-07-25
v1.0 roadmap


valgur:
Although the [zerover](https://0ver.org/) versioning scheme has served us well, I think we are feature-complete enough that it's due time for a proper 1.0 release, especially with the recent additions.
All of the improvement suggestions from an older discussion on this (#108) have now been implemented.

Obviously, all of the outstanding PRs should be merged first, but I would also like to incorporate some other smallish changes/additions:

- [x] 1. I would change the CLI query param format from `-q x=y,z=w` to `-q x=y -q z=w`, which is a bit more flexible, I think.
- [x] 2. Interpret repeated CLI query keywords as a set of allowed values, so for example `-q relativeorbitnumber=105 -q relativeorbitnumber=220` would get translated to `(relativeorbitnumber:105 OR relativeorbitnumber:220)`. This would also apply to `--name` and `--uuid`. (Might as well deprecate them at this point, since they can be replaced by `-q identifier=...` and `-q uuid=...`.) 
- [x] 3. I think the default date range values can cause a lot of confusion, especially if trying to download a specific product with `--name` or `--uuid`. Possible solutions would be to drop the default date range values entirely or to not set it for `--start` and make it a required param. The first one is neater and we kind of provide a warning about excessively large queries by showing a progress bar in such cases. Related: #456
- [x] 4. ~~I would like to add Python API functionality to make downloading individual files from products more convenient. These could be something like `api.list_product_files(uuid, path)`, `api.get_product_file_url(uuid, path)` and `api.download_product_file(uuid, path)`. The list function should probably also include some metadata like file sizes.~~ Actually this can be added after v1.0 release also. Moved to ~~#392~~ #291.
- [x] 5. Should update the readme with the first "getting started" making use of the nice and user friendly `--location` feature instead of `--geometry`.
- [x] 6. Have `--footprints` take output file path. Support for `-` for stdout would also be nice (#153).
- [ ] 7. Convert GeoJSON input into `GEOMETRYCOLLECTION` WKT for multi-feature input. (#114)

Will also have to take another look at the issues collected under the [v1.0 milestone](https://github.com/sentinelsat/sentinelsat/milestone/1) and the milestones of all others.

What do you think? I can probably take care of implementing these suggestions next week, if you agree with them.

valgur:
I updated the milestones for all open issues. Feel free to have a look and adjust them, if necessary. https://github.com/sentinelsat/sentinelsat/milestones

[...]

PR #463, opened 2021-05-08
CLI repeated query param handling

valgur:
Implements the suggestions 1. and 2. in #391:
- Replaces comma-separated strings in `--uuid`, `--name` and `--query` with repeated application of that parameter.
- Repeated query params get interpreted as an "or" expression.


j08lue (approved):
Nice addition with the `key=value` validation callback. 👍 

Great improvement to reduce non-standard CLI behavior before we release v1. Thanks for picking this up!

PR #469, opened 2021-05-14
Use HTTP status code instead of OData metadata for online status

### LINKED ISSUE #467: Invalid checksum/Stream error while trying to download Sentinel-2 files
Ever since the IP update, I have difficulties accessing to the files I could before. I updated sentinelsat to the latest release and added 'https://apihub.copernicus.eu/apihub' to access the SentinelAPI. I also make sure to get data from a month before.

I either get stream error or invalid checksum(corrupted file) errors whatever file I try to access.

Should I wait until they complete their migration or is it a sentinelsat problem?

Thanks.

valgur:
Fixes #467. See https://github.com/sentinelsat/sentinelsat/issues/467#issuecomment-841238253 for details.


j08lue (approved):
This means that every download now incurs an LTA trigger call. That is a bit expensive for a rare issue, but makes the online status much more reliable. This is important, since even more products are offline since the rolling period was shortened to just a month.

So I would not call this a workaround but an improvement of the reliability of online/offline status handling. 😉

PR #441, opened 2021-04-25
New specialized client for the GNSS end-point on scihub

avalentino:
This PR provides a new, specialized, API client for the GNSS end-point on scihub.

~**NOTE**: this PR is based on the ``feature/gha`` that implements CI testing with GitHub Actions.~

Closes: #436.

PR #472, opened 2021-05-15
Avoid guessing of the filename for LTA products

valgur:
For LTA products we currently pretty much guess the filename extension. This is an issue for new or non-standard products like the newly-added GNSS products (#441). This can be made more reliable by fetching the `Filename` attribute from the extended OData metadata and replacing `.SAFE` with `.zip` for the common case.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -59,7 +59,7 @@ Deprecated
 Fixed
 ~~~~~
 * Fix location information for Nominatim bounding box queries (#384)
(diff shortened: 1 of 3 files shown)


avalentino (approved):

j08lue (approved):
Another DHuS detail we rely on here (the existence of `Attributes('Filename')` paths). I hope this works across instances of the system. It is to hope that most of those do not have LTA anyways... Great you knew how to address this!

PR #474, opened 2021-05-16
Use GET instead of POST for queries

### LINKED ISSUE #380: S5P query fails
Impossible to execute query search  on [s5phub portal ](https://s5phub.copernicus.eu/dhus) because the portal doesn't support post requests. 

Replacing session.post with session.get in  _load_subquery() function it works!

It is possible to implement the search as get for sentinel5p?

valgur:
Fixes #380 - the s5phub portal does not accept POST requests in the OpenSearch API.
The only very minor downside is that the search queries are now limited to `latin1` instead of `utf8`.


j08lue (approved):
Well spotted that we could just change to GET requests with the minor consequence of the encoding, which you documented well. 💯

PR #475, opened 2021-05-16
Add --timeout, set default timeout to 60 seconds

### LINKED ISSUE #411: No connection timeout when using CLI
When using the CLI of sentinelsat, there is no option to specify a connection timeout for queries and downloads. There is also no default timeout, which causes the process to hang if the data hub is unavailable.

So it would be nice to have similar functionality as in the Python API to set a connection timeout or at least a default timeout in the CLI.

valgur:
Fixes #411.

I also replaced the default of no timeout with a 60 second limit in both the CLI and API, which seems like a saner default to me. I doubt that most users would prefer to have the client hang indefinitely when the server fails to respond.


j08lue (approved):
👍

PR #477, opened 2021-05-16
Return named tuples for clarity

valgur:


j08lue (approved):
Great, using a named tuple here does add clarity and is very backward-compatible.

PR #479, opened 2021-05-17
Handle cancelled futures

avalentino:
This is connected to #448

### DIFF
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -802,12 +802,11 @@ class SentinelAPI:
 
         failed_prods = {}
         for pid, info in product_infos.items():
-            if (
-                pid not in downloaded_prods
-                and pid not in retrieval_scheduled
-                and not dl_tasks[pid].cancelled()
-            ):
-                info["exception"] = dl_tasks[pid].exception()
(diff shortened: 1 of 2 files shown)


sentinelsat/sentinel.py
```diff
@@ -802,7 +802,11 @@ def download_all(
...
         failed_prods = {}
         for pid, info in product_infos.items():
-            if pid not in downloaded_prods and pid not in retrieval_scheduled:
+            if (
+                pid not in downloaded_prods
+                and pid not in retrieval_scheduled
+                and not dl_tasks[pid].cancelled()
+            ):
                 info["exception"] = dl_tasks[pid].exception()
                 failed_prods[pid] = info
```
valgur, line 811: The cancelled products should still be marked as failed.

```suggestion
            if pid not in downloaded_prods and pid not in retrieval_scheduled:
                if not dl_tasks[pid].cancelled():
                    info["exception"] = dl_tasks[pid].exception()
                failed_prods[pid] = info
```

valgur (changes requested):

valgur:
Good catch, thanks!

PR #480, opened 2021-05-18
Use tqdm.auto

valgur:
The progress indication in notebooks is much-much better as a result.

Before:
Screenshot from 2021-05-18 14-38-05

After:
Screenshot from 2021-05-18 14-36-59


j08lue (approved):
Nice. How long has `tqdm.auto` been around? Should we add that minimum version in the requirements?

valgur:
It's been available since 2018. So quite a while, but I updated the minimum version in requirements.txt nevertheless.

PR #476, opened 2021-05-16
LTA handling improvements

valgur:
- Converted `trigger_offline_retrieval()` into a public method.
- `trigger_offline_retrieval()` now returns True/False depending on whether the product was triggered or is already online and raises an exception in all other cases.
- renamed `SentinelAPILTAError` to `LTAError`, added `LTATriggered` exception.
- `download()` and `get_stream()` now raise `LTATriggered` exception if the product is offline.
- made `get_stream()` more flexible.

PR #486, opened 2021-05-25
Revert to using GET for LTA triggering

valgur:
As reported in https://github.com/sentinelsat/sentinelsat/pull/476#issuecomment-847548274, the change of using HEAD instead of GET for LTA triggering in #457 does not appear to be working as intended.

I replaced the HEAD request with a GET for just a single byte. I tested both the old and new versions manually and the updated version does indeed work.


j08lue (approved):
Cool hack with requesting a single byte! Retains the efficiency of the HEAD request.

PR #490, opened 2021-05-30
Run the weekly live test only in the main sentinelsat/sentinelsat repository

avalentino:
I suppose that the "live-testing.yaml" action was not intended to be run on cloned repositories.
Please consider to add the proposed guard.


valgur:
Oops... :sweat_smile:
Thanks for the fix!

PR #491, opened 2021-06-01
handle 403 error

z4zz:
handle 403 error while Fetching archival status by `download all()`

Full error cause message for online product: 
> An exception occured while creating a stream: Maximum number of 4 concurrent flows achieved by the user "_username_"

Full error cause message for offline product: 
> User '_username_' offline products retrieval quota exceeded (20 fetches max) trying to fetch product _PRODUCT_FILENAME_ (_BYTES_COUNT_ bytes compressed)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,7 @@ CLI changes
 
 Added
 ~~~~~
-* Better handling 403 error in ``trigger_offline_retrieval()`` 
+* Better handling 403 error in `trigger_offline_retrieval()` (`#491 <https://github.com/sentinelsat/sentinelsat/issues/491>`_ `@z4zz <https://github.com/z4zz>`_)
 
 Changed
 ~~~~~~~
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,7 @@ CLI changes
 
 Added
 ~~~~~
-* Better handling 403 error in ``trigger_offline_retrieval()``
+* Better handling 403 error in ``trigger_offline_retrieval()`` 
 
 Changed
 ~~~~~~~
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,7 @@ CLI changes
 
 Added
 ~~~~~
-* 
+* Better handling 403 error in ``trigger_offline_retrieval()``
 
 Changed
 ~~~~~~~
--- a/sentinelsat/sentinel.py
+++ b/sentinelsat/sentinel.py
@@ -630,10 +630,12 @@ class SentinelAPI:
         elif r.status_code == 202:
             self.logger.debug("Accepted for retrieval")
             return True
-        elif r.status_code == 403 and "concurrent flows" in cause:
-            self.logger.debug("Product is online")
+        elif r.status_code == 403 and cause and "concurrent flows" in cause:
+            # cause: 'An exception occured while creating a stream: Maximum number of 4 concurrent flows achieved by the user "username""'
+            self.logger.debug("Product is online but concurrent downloads limit was exceeded")
             return False
         elif r.status_code == 403:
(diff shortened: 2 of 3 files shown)


sentinelsat/sentinel.py
```diff
@@ -630,6 +630,9 @@ def trigger_offline_retrieval(self, uuid):
         elif r.status_code == 202:
             self.logger.debug("Accepted for retrieval")
             return True
+        elif r.status_code == 403 and "concurrent flows" in cause:
+            self.logger.debug("Product is online")
+            return False
         elif r.status_code == 403:
             msg = f"User quota exceeded: {cause}"
             self.logger.error(msg)
```
valgur, line 638: I would also include the full cause messages as code comments for context.
```suggestion
        elif r.status_code == 403 and cause and "concurrent flows" in cause:
            # cause: 'An exception occured while creating a stream: Maximum number of 4 concurrent flows achieved by the user "username""'
            self.logger.debug("Product is online but concurrent downloads limit was exceeded")
            return False
        elif r.status_code == 403:
            # cause: 'User 'username' offline products retrieval quota exceeded (20 fetches max) trying to fetch product PRODUCT_FILENAME (BYTES_COUNT bytes compressed)'
            msg = f"User quota exceeded: {cause}"
            self.logger.error(msg)
```

valgur (changes requested):
Good catch, thanks!
We should also add a simple test case for it here https://github.com/sentinelsat/sentinelsat/blob/main/tests/test_download.py#L75-L97
Do you want to take a shot at it or I could also take care of it in a separate PR?

z4zz:
I'm not familiar with tests, I'll leave that to you

j08lue:
> I'm not familiar with tests, I'll leave that to you

Maybe now is the time to get into it, @z4zz? 😉

j08lue:
@z4zz something went wrong with CI. How about you add a CHANGELOG entry for this fix, which should trigger it cleanly (and is nice to get, anyways)?

j08lue:
If you just add a line to CHANGELOG, it might go well from there. If not, we should raise another issue to fix the Autoblack/Pytest interplay.

tests/test_download.py
```diff
@@ -53,14 +53,22 @@ def test_dhus_version(dhus_url, version):
...
+        (206, False, {}),
         # Accepted for retrieval - the product offline product will be retrieved from the LTA.
-        (202, True),
+        (202, True, {}),
+        # already online but concurrent downloads limit was exceeded
+        (
+            403,
+            False,
+            {
+                "cause-message":'An exception occured while creating a stream: Maximum number of 4 concurrent flows achieved by the user "mock_user"'}
```
j08lue, line 69:
```suggestion
                "cause-message":'An exception occured while creating a stream: Maximum number of 4 concurrent flows achieved by the user "mock_user"'
```

CHANGELOG.rst
```diff
@@ -12,7 +12,7 @@ CLI changes
 
 Added
 ~~~~~
-* 
+* Better handling 403 error in ``trigger_offline_retrieval()``
```
j08lue, line 15:
```suggestion
* Better handling 403 error in `trigger_offline_retrieval()` (`#491 <https://github.com/sentinelsat/sentinelsat/issues/491>`_ `@z4zz <https://github.com/z4zz>`_)
```

j08lue:
Great you added the test @z4zz. There was a syntax error blocking the CI, but that is fixed now. Now there is only a bit of test logic that needs fixing, it seems, before this PR can go through. @valgur, could you be persuaded to look into that?

j08lue:
Ok, never mind. I get what is happening: In some cases, the server includes a `cause-message` header when it returns an error code (in this case 403), and you use that to distinguish between different cases of 403. That header is what you added to the tests together with a new case, in order to distinguish between the cause being too many streams for an online product or too many fetch requests for an LTA product. What a mess this LTA has caused...

But nice catch, thanks @z4zz, we can merge this now.

Issue #493, opened 2021-06-06
HTTP status 403 Forbidden: User quota exceeded


ibaris:
The `download_all` module causes the following error message:

```python
Fetching archival status:  17%|█▋        | 4/24 [00:05<00:27,  1.36s/ products]User quota exceeded: MediaRegulationException : An exception occured while creating a stream: Maximum number of 4 concurrent flows achieved by the user "xxx"
Fetching archival status:  17%|█▋        | 4/24 [00:06<00:34,  1.72s/ products]
Traceback (most recent call last):
  File "C:\Users\xxx\miniforge3\lib\site-packages\IPython\core\interactiveshell.py", line 3437, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-7-35ca78a4b9cf>", line 1, in <module>
    __api.download_all(product_ids, directory_path=dest_path, n_concurrent_dl=1)
  File "C:\Users\xxx\miniforge3\lib\site-packages\sentinelsat\sentinel.py", line 723, in download_all
    is_online = not self.trigger_offline_retrieval(pid)
  File "C:\Users\xxx\miniforge3\lib\site-packages\sentinelsat\sentinel.py", line 636, in trigger_offline_retrieval
    raise LTAError(msg, r)
sentinelsat.exceptions.LTAError: HTTP status 403 Forbidden: User quota exceeded: MediaRegulationException : An exception occured while creating a stream: Maximum number of 4 concurrent flows achieved by the user "xxx"
```

Even if I try to set the parameter `n_concurrent_dl` to `1`, the result is the same. However, I am able to download every single product with a loop and by using  the `download` function:

```python
for item in products.keys():
    api.download(item, "")
```

Why I wanted to use the `download_all ` function is, I would like to download conveniently products from the LTA, as you described [here](https://sentinelsat.readthedocs.io/en/master/api_overview.html#lta-products). 

Are there a workaround for this issue? I saw the merge request  #491. But even with this code the `download_all` function does not work for me.

z4zz:
the main reason is that online status is now checked by triggering function 
you have 2 options:
1. modify `download_all` (line **723**): `is_online = not self.trigger_offline_retrieval(pid)` to `is_online = self.is_online(pid)`
2. modify `trigger_offline_retrieval` (line **636**): `raise LTAError(msg, r)` to `return False`

ibaris:
[...]

PR #508, opened 2021-07-03
Improve handling of offline products

avalentino:
Never try to trigger LTA retrieval if products are online.


valgur:
Using `is_online = not self.trigger_offline_retrieval(id)` instead of `is_online = self.is_online(id)` is very deliberate. The `Online` property of products has unfortunately not been 100% reliable and testing with a `trigger_offline_retrieval()` is a foolproof way to get the true status of the product. It's an extra query but is negligible considering the typical download size.

avalentino:
> Using is_online = not self.trigger_offline_retrieval(id) instead of is_online = self.is_online(id) is very deliberate. The Online property of products has unfortunately not been 100% reliable and testing with a trigger_offline_retrieval() is a foolproof way to get the true status of the product. It's an extra query but is negligible considering the typical download size.

The problem is that I often get the following error message

```
User quota exceeded: ExpectedException : User '****' offline products retrieval quota exceeded (20 fetches max)  trying to fetch product S1A_S3_GRDH_1SDV_20201214T152857_20201214T152916_035683_042CC9_BDDA (131284702 bytes compressed)
```

even in the case most of the requested products are online and few of the are offline.
My feeling (not 100% sure) is that the use of `self.trigger_offline_retrieval(id)` seems to increase in any case the count of retrieval attempts even if products are actually online.

Do you think there could be some kind of alternative strategy?

j08lue:
> My feeling (not 100% sure) is that the use of `self.trigger_offline_retrieval(id)` seems to increase in any case the count of retrieval attempts even if products are actually online.

Geez, they are not making this easy! Would you perhaps ask EO Support, whether these status requests indeed count against the quota? I guess this is happening in some kind of API gateway (I hope!), but might still be something they can configure?

Otherwise I would too advocate for going back to the old way of relying on the `Online` status being correct and failing when it is not.

avalentino:
> Geez, they are not making this easy! Would you perhaps ask EO Support, @avalentino, whether these status requests indeed count against the quota? I guess this is happening in some kind of API gateway (I hope!), but might still be something they can configure?

Unfortunately this overlaps with #493 so it id hard to provide a reproducible example to the support team.
I will try in any case.

j08lue:
I also saw the issue with tons of 403 errors (too many requests) from `api.download_all` and the change in this branch got rid of them. Curiously, I got these issues only when downloading Sentinel-1 SLC products (and not even LTA ones).

I propose to merge this and open an issue to go back to calling `trigger_offline_retrieval` to reliably get the online status. What do you think, @valgur?

valgur:
I agree, let's merge this change. Will have to think of a solution for dealing with incorrect Online status in OData, but that's a much-much less severe issue and can be ignored in the short term.

PR #511, opened 2021-07-04
Live testing fixes

valgur:
The live test have been failing for the past month and more. This has been caused by two broad server-side changes:
* products get archived after just 30 days, making the set of usable products for testing change constantly;
* the smallest online products tend to be ones affected by the data corruption bug during retrieval from archive.

This PR makes the tests more robust against these two issues.


j08lue (approved):
Working with SciHub has turned into a fairytale from 1001 nights - so many riddles to solve... You still prevail, @valgur, great work. 😸

Issue #521, opened 2021-07-21
Handle query length limits automatically


daniel-j-h:
Hey folks,

I'm on sentinelsat 1.0.1 sending geojson with the `geojson_to_wkt` function.

Some geojson objects seem to be too big in their WKT encoding for the server, and I'm
1. getting a 500 exception from the server
2. the server sending me the full java backtrace (?)

We know the server limit in sentinelsat, no? Can we either
- check the length automatically before we send the query to the server
- somehow notify the copernicus folks to not run into a 500 and not send the exception backtrace
- automatically turn geometries into their bounding box and search for the bounding box

I'm not 100% sure if querying for the geom's bounding box will always return the same results - thoughts?

Here is the full log.

```
Traceback (most recent call last):
  File "/home/python/.local/lib/python3.8/site-packages/sentinelsat/sentinel.py", line 1433, in _check_scihub_response
    response.raise_for_status()
  File "/home/python/.local/lib/python3.8/site-packages/requests/models.py", line 953, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: https://apihub.copernicus.eu/apihub/search?<radacted>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  <redacted>
  File "/home/python/.local/lib/python3.8/site-packages/sentinelsat/sentinel.py", line 216, in query
    response, count = self._load_query(query, formatted_order_by, limit, offset)
  File "/home/python/.local/lib/python3.8/site-packages/sentinelsat/sentinel.py", line 289, in _load_query
    products, count = self._load_subquery(query, order_by, limit, offset)
  File "/home/python/.local/lib/python3.8/site-packages/sentinelsat/sentinel.py", line 322, in _load_subquery
    _check_scihub_response(response, query_string=query)
  File "/home/python/.local/lib/python3.8/site-packages/sentinelsat/sentinel.py", line 1479, in _check_scihub_response
    raise ServerError(msg, response)
sentinelsat.exceptions.ServerError: HTTP status 500 Internal Server Error:
# HTTP Status 500 – Internal Server Error

* * *

 **Type** Exception Report

 **Message** Could not roll back Hibernate transaction; nested exception is
org.hibernate.TransactionException: Unable to rollback against JDBC Connection

 **Description** The server encountered an unexpected condition that prevented
it from fulfilling the request.

 **Exception**



[...]

PR #525, opened 2021-07-22
Suppress unnecessary server stacktrace printing for QueryLengthError

valgur:
As suggested in #521.


j08lue (approved):
Nice! Get rid of all that server mess... Gotta love how it only required a small addition in a single line.

PR #526, opened 2021-07-23
New CLI option to specify the format of the output report

avalentino:


valgur:
That's a very nice addition. Thanks!

Just one nitpick: the attribute list in the documentation is specific to Sentinel-1. And even if it covered other Sentinels, these might still not be applicable to products from other portals. I would instead perhaps instruct the user to run a relevant query with `--footprints -` (plus maybe `--limit 1`) and use the attributes that are available there (plus `footprint` and `gmlfootprint` that are not included in the GeoJSON output).

avalentino:
Thanks @valgur 

> Just one nitpick: the attribute list in the documentation is specific to Sentinel-1. And even if it covered other Sentinels, these might still not be applicable to products from other portals.

good point

> I would instead perhaps instruct the user to run a relevant query with --footprints - (plus maybe --limit 1) and use the attributes that are available there (plus footprint and gmlfootprint that are not included in the GeoJSON output).

OK, I have added the instructions to get the full list.
Also think I was able to reduce the list of keys to a common sub-set (I was not able to check Sentinel-5).
Anyway if you prefer I can remove the list totally.

sentinelsat/scripts/cli.py
```diff
@@ -185,6 +185,24 @@ def validate_query_param(ctx, param, kwargs):
...
+    show_default=True,
+    help="""Specify a custom format to print results. The format string shall
+    be compatible with the Python "Format Specification Mini-Language".
+    Possible keys for substitution are: 'beginposition', 'endposition',
+    'filename', 'format', 'id', 'identifier', 'ingestiondate',
+    'instrumentname', 'instrumentshortname', 'link', 'link_alternative',
+    'link_icon', 'ondemand', 'orbitdirection', 'orbitnumber',
+    'platformidentifier', 'platformname', 'producttype',
+    'relativeorbitnumber', 'size', 'summary', 'title', 'uuid',
+    'footprint', 'gmlfootprint'.
```
j08lue, line 200: I would also suggest to remove this list, since it is likely to change.
```suggestion
```

docs/cli.rst
```diff
@@ -204,6 +204,25 @@ Options
...
+    be compatible with the Python "Format Specification Mini-Language".
+
+    Possible keys for substitution are: 'beginposition', 'endposition',
+    'filename', 'format', 'id', 'identifier', 'ingestiondate',
+    'instrumentname', 'instrumentshortname', 'link', 'link_alternative',
+    'link_icon', 'ondemand', 'orbitdirection', 'orbitnumber',
+    'platformidentifier', 'platformname', 'producttype',
+    'relativeorbitnumber', 'size', 'summary', 'title', 'uuid',
+    'footprint', 'gmlfootprint'.
+
```
j08lue, line 219:
```suggestion
```

j08lue (changes requested):
Nice idea to make this configurable. I think the list of possible keys should not be included, as this varies a lot depending on sensor, DHuS, etc.

avalentino:
Thanks @j08lue for the review.
I have updated the code according to your requests.

j08lue (approved):

docs/cli.rst
```diff
@@ -204,6 +204,17 @@ Options
...
     Query orbit products form the GNSS end-point ("https://scihub.copernicus.eu/gnss").
 
+.. option:: --fmt <format string>
+
+    Specify a custom format to print results. The format string shall
+    be compatible with the Python "Format Specification Mini-Language".
+
+    A complete list keys that can be used for a specific sentinel mission
+    can be e.g. obtained looking at the "properties" of the output produced by
+    the ``--footprints`` flag  - (maybe with ``--limit 1``).
```
valgur, line 214: Should include the dash in `--footprints -`, which directs the output to stdout. I would still also include a subjective list of common and likely useful keywords as well. Suggested wording:
```suggestion
    Some common keywords for substitution are:
      'uuid', 'title', 'summary', 'link', 'size', 'platformname', 'producttype',
      'beginposition', 'instrumentshortname', 'cloudcoverpercentage',
      'orbitdirection', 'relativeorbitnumber', 'footprint'.

    For a complete set of available keywords see the "properties" output from a relevant query with ``--footprints -`` (and possibly ``--limit 1``) appended.
```

valgur (changes requested):

docs/cli.rst
avalentino: @valgur your suggested wording is OK for me, just two questions:

* what about the ``identifier`` and/or ``filename`` keys? These are the primary reason why I needed to add this feature. Do you think that they can be added to the list?
* you suggested a rewording for the sphinx documentation. The CLI help message and the README shall be updated accordingly?

@j08lue the above changes are OK for you? Do you have any comment?

valgur: `identifier` is identical to `title`. I suppose `identifier` is a bit clearer, so you can replace it with that. Feel free to add `filename` as well.
I would add the more detailed info to the CLI help message, too, sure, but not to the README.

avalentino: ok, thanks

valgur (approved):

j08lue (approved):

PR #531, opened 2021-07-31
Use tqdm.write for nicer logging + progress bars

valgur:
This ensures that progressbars don't overwrite logging output and vice versa.
However, had to include a `DISABLE_TQDM_LOGGING` environment variable option to not break tests.


avalentino (approved):

j08lue (approved):
Nice CLI improvement! A bit much boilerplate code to add for such a simple feature, for my taste, but ok when it is just in the outer interface layer.

PR #541, opened 2021-09-16
Fix filter_node function (closes #540)

### LINKED ISSUE #540: Download TCI file only
Hello,

I am trying to download only the TCI file using `sentinelsat`. I use the following code:

`api = SentinelAPI('username', 'password')`
`path_filter = make_path_filter("*_TCI_10m.jp2")`
`api.download('c8579d0d-07ee-4b79-861d-4abb87672659', nodefilter=path_filter)`

and I got the product ID from an `api.query()`. However, only the `manifest.safe` is downloaded and not the TCI file.

Any idea what I am doing wrong?

Thanks in advance,

Zeno

avalentino:
Do not convert the node path to lowercase before pattern matching.


valgur:
Thanks! Please also add a changelog entry for this fix.

PR #550, opened 2021-11-22
Drop Gitter

valgur:
I propose we drop the Gitter link from the readme. I don't think any maintainer is really keeping an eye on it with any regularity, if at all. The issue tracker appears to be quite sufficient for assistance now and we could also consider enabling the Discussions page feature.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,10 @@
 sentinelsat
 ===========
 
+.. image:: https://readthedocs.org/projects/sentinelsat/badge/?version=stable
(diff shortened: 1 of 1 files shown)


j08lue (approved):
Yes, thanks.

Fernerkundung:
+1 - it has seen very little use in the past as well and the issue tracker was much more active for users of sentinelsat helping each other.

Fernerkundung (approved):

PR #565, opened 2022-06-06
Pin rstcheck to v5, fix tests on Python 3.10

valgur:
The `rstcheck` API has changed significantly and the Python API has been moved to `rstcheck-core`.

However, the newer versions no longer support Python 3.6 and I don't find this change to be a good enough of a reason for us to drop Python 3.6 support, so I am pinning `rstcheck` to an older version instead. The output from the newer version was identical to the older one's.

Also added Python 3.10 to CI and fixed a minor related test incompatibility.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -8,7 +8,7 @@ from functools import partialmethod
 
 try:
     from test.support.os_helper import EnvironmentVarGuard
-except:
+except ImportError:
     from test.support import EnvironmentVarGuard
 
 import pytest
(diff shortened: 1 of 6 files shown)


tests/test_cli.py
```diff
@@ -5,7 +5,11 @@
 import shutil
 from contextlib import contextmanager
 from functools import partialmethod
-from test.support import EnvironmentVarGuard
+
+try:
+    from test.support.os_helper import EnvironmentVarGuard
+except:
```
j08lue, line 11:
```suggestion
except ImportError:
```

j08lue (approved):
Great, thanks for the good house-keeping.

Issue #566, opened 2022-07-03
How to download only enchanced_measrement.nc file from a Sen-3 SRAL product using node filter ?


usamarehan557:
Hi,

I'm trying to download only enhanced_measurement.nc file from a SEN 3 SRAL product using node_filter but I'm getting error every time:

**Unexpected nav segment Navigation Property: org.apache.olingo.odata2.core...**

I think I'm not making my node_filter right so kindly help me make a node_filter for my case.

Thanks alot.

avalentino:
Hi @usamarehan557, could you please provide more information?
A complete example reproducing the issue and a complete error traceback would surely help.

usamarehan557:
`def print_products():
    api = SentinelAPI(username, password, 'https://apihub.copernicus.eu/apihub')
    sukkur_footprint = geojson_to_wkt(read_geojson('sukkur_search_polygon.geojson'))

    products = api.query(
        sukkur_footprint,
        relativeorbitnumber=105,
        date=('20220201', '20220228'),
        platformname='Sentinel-3',
        filename='S3A_*',
        producttype='SR_2_LAN___',
        timeliness='Non Time Critical',
        instrumentshortname='SRAL',
        productlevel='L2')


    # print(products)

    if products:
        geojson = api.to_geojson(products)
        # geodataframe = api.to_geodataframe(products)
        properties = geojson.features[0].properties
        uuid = properties['uuid']
        title = properties['title']
        file_path = '*measurement/*'
        path_filter = make_path_filter(file_path)
        product_info = api.get_product_odata(properties['uuid'])
        is_online = product_info['Online']

        if is_online:
            print(f'Product {uuid} is online. Starting download. Title : {title}.SEN3')
            # print(geodataframe)
            try:
                api.download(uuid, nodefilter=path_filter)
                # requests.get('https://scihub.copernicus.eu/dhus/odata/v1/Products('{uuid}')/Nodes('{title}')/Nodes(
                # 'enhanced_measurement.nc')/$value', headers={})
            except Exception as e:
                print(e)
        else:
            print(f'Product {uuid} is not online.')
            # api.trigger_offline_retrieval(uuid)
    else:
        print('Products not found')`

Error Log: 

`Product a8cf9e3f-f91e-473e-9067-811969feda7f is online. Starting download. Title : S3A_SR_2_LAN____20220217T052540_20220217T061609_20220314T222335_3029_082_105______LN3_O_NT_004.SEN3
[...]

PR #573, opened 2022-09-23
Fix nodefilter not working with Sentinel-3 products

valpamp:
Fixes issue #566, which was caused by the code not handling the product folder path and manifest filename based on the type of product. Since most Sentinel products have a folder path ending in .SAFE and a manifest file named manifest.safe, this was an issue only with products such as Sentinel-3, whose folder path ends in .SEN3 and whose manifest file is named xfdumanifest.xml.
By fetching the full odata metadata of the queried product, it is possible to get the correct folder path, and sets the manifest filename accordingly.

If there are products whose folder path does not end in '.SAFE' or '.SEN3', they should be handled accordingly.


avalentino (approved):
LGTM

j08lue:
I agree, looks good and is an improvement in terms of avoiding hard-coded names. 

CI shows you need to Black-format, please, then we'll know how the rest of the tests go.

valpamp:
> CI shows you need to Black-format, please, then we'll know how the rest of the tests go.

I used the black linter included in Spyder and the number of changes in the code is much higher than I expected. From what I can see basically all of them seem to be breaking down lines that are longer than 79 characters. Is this ok? Or did I use the wrong code formatter?

avalentino:
@valpamp the configuration for black is in the `pyproject.toml` file

https://github.com/sentinelsat/sentinelsat/blob/3682e436f2cc2704ca8e93579eafb09cad8874f7/pyproject.toml#L1-L16

The `line-length` is supposed to be 100.

valpamp:
I finally managed to run black with using the proper configuration file, it should be ok now.

j08lue:
Great, now Black is happy, but one test is failing because the product info we expect (from recordings) is different from what we have now, where the `folder_path` and `manifest` properties are added. https://github.com/sentinelsat/sentinelsat/actions/runs/3134363937/jobs/5090260771#step:8:80

Would you try to re-record that cassette, @valpamp? There are some instructions on that here: https://github.com/sentinelsat/sentinelsat/#tests If it does not work for you, we can help.

Before you do, I am not super happy with the naming of `folder_path`. It is more like a filename - that, for unzipped products, is more like a folder name. You dub it `filename` yourself here:
https://github.com/sentinelsat/sentinelsat/pull/573/files#diff-8539d03b31a9a42eeab364c8eafc747a91bef19778c6cd4fc3630488c1fc7439R1075

So should we still consider renaming it to `target_filename` or `product_name` or something? I think `folder_path` is perhaps a bit too generic for what this property is.

valpamp:
`product_name` sounds good to me, I will make sure to rename it. Regarding the test, I have skimmed through the vcrpy documentation, and if I understand correctly I should just run pytest on the failing test using the `--vcr-record all` argument to force the cassette to be re-recorded, and if everything goes right I should see the associated yaml file change in the vcr_cassettes folder.

j08lue:
Great, thanks, @valpamp! Sounds right with `--vcr-record all`.

valpamp:
Hey everyone, sorry this took so long, I had to turn in my PhD thesis and I have been completely absorbed in that for a while.

Today I finally ran pytest in my conda environment, and the test failure seems to be caused by a mismatch between the content of the `product_node` set and the second string:

https://github.com/sentinelsat/sentinelsat/blob/7fd4122cdf85df651ba41d843ae17a6697106483/tests/test_download.py#L376

The `product_node` set contains the string `'preview/map-overlay.kml'`, but it is being checked against `'./preview/map-overlay.kml'`. Removing the initial `'./'` from the control string and running pytest with `--vcr-record all` apparently solves the issues, and the cassettes are re-recorded successfully. 

However, if I now run pytest **without** the `--vcr-record all` option, `test_download_nodes `fails because of an xml parsing error. The culprit is the manifest.SAFE file of the S1A_IW_GRDH_1SDV_20210722T182826_20210722T182851_038894_0496E2_5BD9 product being downloaded during the test: checking the manifest.SAFE file, I can clearly see that it ends abruptly at the line where the xml parsing library complains, but I have no idea why. Any suggestions?

j08lue:
I can see the test is using the first of these recent Sentinel-1 products:

https://github.com/sentinelsat/sentinelsat/blob/7fd4122cdf85df651ba41d843ae17a6697106483/tests/conftest.py#L293
https://github.com/sentinelsat/sentinelsat/blob/7fd4122cdf85df651ba41d843ae17a6697106483/tests/test_download.py#L366

Have you tried using a different one, e.g. the second from that query? Or try again tomorrow, when the query hopefully returns some different products. If the problem persists, it might be more systematic.

Thanks for your efforts - and good luck with the thesis!

valpamp:
I ran more tests with the other two products and I am getting different errors. If I run the test with `--vcr-record all` everything works fine in both cases, but then if I re-run the test without the option I get a `CannotOverwriteExistingCassette` error, which does not make much sense to me since I just re-recorded the cassette while pointing to the same product. But maybe it's just me having no idea how VCR works 😅 
I also tried seeing if something changed with the first product, but the parsing error persists for now, the xml still appears to be truncated.

Finally, yesterday I found and solved another weird issue with the node filter on Sentinel-2 products from 2018. But I will open another issue/PR for that.

j08lue:
I can reproduce the issue with the truncated manifest XML:

1. Run `pytest -v --record-vcr all tests/test_download.py::test_download_product_nodes`
2. Run `pytest -v tests/test_download.py::test_download_product_nodes`

```
______________________________________________________________________________________________________________________________________________________________________ test_download_product_nodes _______________________________________________________________________________________________________________________________________________________________________

api = <sentinelsat.sentinel.SentinelAPI object at 0x11012bdc0>, tmpdir = local('/private/var/folders/fj/h7p8zmrs5pv538jqx4b0z7c80000gn/T/pytest-of-jonassolvsteen/pytest-8/test_download_product_nodes0')
node_test_products = [{'Creation Date': datetime.datetime(2021, 7, 22, 19, 35, 3, 413000), 'Ingestion Date': datetime.datetime(2021, 7, 22,...ime(2021, 7, 22, 19, 24, 14, 449000), 'Online': True, 'date': datetime.datetime(2021, 7, 22, 17, 43, 56, 232000), ...}]

    @pytest.mark.vcr
    @pytest.mark.scihub
    def test_download_product_nodes(api, tmpdir, node_test_products):
        uuid = node_test_products[0]["id"]
        product_dir = node_test_products[0]["title"] + ".SAFE"
        expected_path = tmpdir.join(product_dir)
    
        nodefilter = make_path_filter("*preview/*.kml")
>       product_info = api.download(uuid, str(tmpdir), checksum=True, nodefilter=nodefilter)

tests/test_download.py:371: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
sentinelsat/sentinel.py:590: in download
    return downloader.download(id, directory_path)
sentinelsat/download.py:133: in download
    return self._download_with_node_filter(id, directory, stop_event)
sentinelsat/download.py:164: in _download_with_node_filter
    node_infos = self._filter_nodes(manifest_path, product_info, self.node_filter)
sentinelsat/download.py:779: in _filter_nodes
    xmldoc = etree.parse(manifest)
/opt/homebrew/Cellar/python@3.10/3.10.6_2/Frameworks/Python.framework/Versions/3.10/lib/python3.10/xml/etree/ElementTree.py:1222: in parse
    tree.parse(source, parser)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <xml.etree.ElementTree.ElementTree object at 0x1101293c0>, source = <_io.BufferedReader name='/private/var/folders/fj/h7p8zmrs5pv538jqx4b0z7c80000gn/T/pytest-of-jonassolvsteen/pytest-8/test_download_product_nodes0/S1A_IW_GRDH_1SDV_20210722T182826_20210722T182851_038894_0496E2_5BD9.SAFE/manifest.safe'>
parser = <xml.etree.ElementTree.XMLParser object at 0x10773f520>

    def parse(self, source, parser=None):
        """Load external XML document into element tree.
    
        *source* is a file name or file object, *parser* is an optional parser
        instance that defaults to XMLParser.
    
        ParseError is raised if the parser fails to parse the document.
    
        Returns the root element of the given source document.
    
        """
        close_source = False
        if not hasattr(source, "read"):
            source = open(source, "rb")
            close_source = True
        try:
            if parser is None:
                # If no parser was specified, create a default XMLParser
                parser = XMLParser()
                if hasattr(parser, '_parse_whole'):
                    # The default XMLParser, when it comes from an accelerator,
                    # can define an internal _parse_whole API for efficiency.
                    # It can be used to parse the whole source without feeding
                    # it with chunks.
>                   self._root = parser._parse_whole(source)
E                   xml.etree.ElementTree.ParseError: unclosed token: line 292, column 69

/opt/homebrew/Cellar/python@3.10/3.10.6_2/Frameworks/Python.framework/Versions/3.10/lib/python3.10/xml/etree/ElementTree.py:580: ParseError
```

The file `S1A_IW_GRDH_1SDV_20210722T182826_20210722T182851_038894_0496E2_5BD9.SAFE/manifest.safe` is truncated after 292 lines.

Have you tried downloading that product separately and seeing whether the file is complete then?

j08lue:
The recorded fixture `tests/fixtures/vcr_cassettes/data/manifest.safe` appears to be complete (332 lines).

j08lue:
Btw, why is the `manifest.safe` even downloaded when the node filter pattern in the test does not seem to include it: `"*preview/*.kml"`?

https://github.com/sentinelsat/sentinelsat/blob/821122437ce224a08323c2c4093ef6f2a23d38ab/tests/test_download.py#L370-L371

avalentino:
@j08lue the manifest contains the full list of files included in the product. It is always downloaded and parsed to extract the file list (and associate metadata), then filters use this information to decide if a file shall be actually downloaded.

valgur (approved):

j08lue (approved):
Thanks for completing this long-standing PR, @valgur.

PR #612, opened 2024-02-28
skip non-standard content in osearch response

ninsbl:
Sentinel-3 products from the Norwegian Ground segment for example cannot be parsed by `sentinelsat` _1.2.1_ because the response contains un-expected, non-standard elements and `_parse_opensearch_response` fails if the response contains such non-standard items.

This PR adds two lines to skip these cases...


sentinelsat/sentinel.py
```diff
@@ -1384,6 +1384,8 @@ def default_converter(x):
                             name = "link_" + p["rel"]
                         product_dict[name] = p["href"]
                 else:
+                    if "content" not in p and "str" not in p:
+                        continue
```
j08lue: Was this really the line where you wanted to add this check? 

What is `p` supposed to be at this point, since it is only defined in the first branch on line 1381 (`for p in properties`)?

j08lue (changes requested):
Great to fix this for the Norwegian ground segment.

I think the code is not correct, though, please see my comment below.

Can you add a test for this condition, too? Could be a simple unit test for `_parse_opensearch_response` like this one,

https://github.com/ninsbl/sentinelsat/blob/4cbff773b4d450741668e26de369f9c25a50a6a8/tests/test_misc.py#L122-L131

just with a blob of representative sample data from the Norwegian ground segment.

sentinelsat/sentinel.py
j08lue: I see you already fixed it.

ninsbl:
Thanks for the swift reply. Happy to add a unittest. I just wonder in the example, where does the _raw_products_ input here: https://github.com/ninsbl/sentinelsat/blob/4cbff773b4d450741668e26de369f9c25a50a6a8/tests/test_misc.py#L123 come from?

ninsbl:
Failing tests seem unrelated (e.g. black check does not pass for unchanged test files with Pythpn > 3.7).

The new, added test passes:
https://github.com/sentinelsat/sentinelsat/actions/runs/8089096858/job/22112812013?pr=612#step:8:100

j08lue (approved):
Thanks for adding the test, nice improvement that keeps this library relevant to some ground segments that continue to use DHuS.

PR #618, opened 2024-09-30
CI: drop Python 3.7, add 3.12

valgur:
Python 3.7 is no longer available on Windows GitHub runners and is not that relevant overall anymore.


j08lue (approved):
👌

